PR #16, opened 2024-10-23
feat: Feature gate evm dependencies

kayibal:
Currently we still rely on ethers in some core models - we can live with this for now. 

There are some other modules that seem more general purpose and ideally we should avoid requiring a feature gate around them:

- VM (not quite sure what this is for tbh, but since it is only VM it sounds general purpose)
- Tycho (we do some decoding here - this should probably be just moved into one of the already gated modules)

EDIT: been rebased on the latest changes. Tycho decoding has been moved to the individual pool states already, it is just VM that still needs to be looked into.


dianacarvalho1 (approved):
Thank you Thank you @louise-poole 🙏🏼

PR #27, opened 2024-10-30
feat: (VMPoolState) implement get_amount_out

tamaralipows:
**Major fixes implemented:**
- Initialize code for DAI contract in test initialization
- Properly overwrite DAI balance and allowance slots for our EOA and Adapter contracts
- Change swap method to interpret Token::Tuple result like the simulation returns, not array (this was breaking).

**TODO:**
- [x]  This currently implements the inner `_get_amount_out` which doesn't check for the `HardLimits` capability - this needs to be added
- [x]  Add test `get_amount_out_sell_limit`
- [x]  Add test `get_amount_out_dust` 
- [x]  Return `GetAmountOut` result from `get_amount_out`, including updated pool state
- [x]  Move get_amount_out into separate `ProtocolSim` Trait
- [x] CI fails because of the Tycho tutorial. Waiting for someone to fix this lol.
- [x] Deal with overwrites - I (Diana) still don't know what to do here 😩 
- [x] Simulation methods don't actually need to be async

**Note:** The current test_get_amount_out is more or less equivalent to `test_get_amount_out` in `test_third_party_pool.py` - just with the sell and buy tokens switched.

For the next PR:
- Figure out if we do need the extra lock or not


src/protocol/vm/state.rs
```diff
@@ -533,28 +536,115 @@ impl VMPoolState<PreCachedDB> {
...
+        if new_price != 0.0f64 {
+            new_state
+                .marginal_prices
+                .insert((sell_token.address, buy_token.address), new_price);
+            new_state
+                .marginal_prices
+                .insert((buy_token.address, sell_token.address), 1.0f64 / new_price);
+        }
+
+        // Convert the received amount to buy token units
```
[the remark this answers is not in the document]
tamaralipows: fml lol

src/protocol/vm/state.rs
```diff
@@ -589,12 +679,45 @@ mod tests {
...
...1526020819052604090209082900390555b6001600160a01b03821661055657600280548290039055610574565b6001600160a01b0382165f9081526020819052604090208054820190555b816001600160a01b0316836001600160a01b03167fddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef836040516103cf91815260200190565b5f6020808352835180828501525f5b818110156105e4578581018301518582016040015282016105c8565b505f604082860101526040601f19601f8301168501019250505092915050565b80356001600160a01b038116811461061a575f80fd5b919050565b5f8060408385031215610630575f80fd5b61063983610604565b946020939093013593505050565b5f805f60608486031215610659575f80fd5b61066284610604565b925061067060208501610604565b9150604084013590509250925092565b5f60208284031215610690575f80fd5b61069982610604565b9392505050565b5f80604083850312156106b1575f80fd5b6106ba83610604565b91506106c860208401610604565b90509250929050565b600181811c908216806106e557607f821691505b60208210810361070357634e487b7160e01b5f52602260045260245ffd5b50919050565b8082018082111561023e57634e487b7160e01b5f52601160045260245ffdfea2646970667358221220dfc123d5852c9246ea16b645b377b4436e2f778438195cc6d6c435e8c73a20e764736f6c63430008140033000000000000000000000000000000000000000000000000000000000000000000")?);
```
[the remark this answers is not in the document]
tamaralipows: I don't see our setup properly initializing DAI's ERC20 code anywhere?
I was getting `AddressEmptyCode(address)` which is coherent with the code not being there. I'll double check though

src/protocol/vm/state.rs
```diff
@@ -533,28 +536,115 @@ impl VMPoolState<PreCachedDB> {
...
+        sell_amount: U256,
+        buy_token: ERC20Token,
+    ) -> Result<(U256, U256, VMPoolState<PreCachedDB>), ProtosimError> {
+        let sell_amount_limit = self
+            .clone()
+            .get_sell_amount_limit(vec![sell_token.clone(), buy_token.clone()])
+            .await?;
+
+        let overwrites = self
+            .clone()
```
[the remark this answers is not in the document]
tamaralipows: Repeating my private message to you...

I don’t know what to do about the block lasting overwrites - since I can’t get rid of the self.clone without changing the entire interface of get_amount_out to take &mut self , which would affect even the native protocols. Seems like a big no-no to me, since we have already released the native protocols I guess? I’ve started looking into removing block_lasting_overwrites, but I think I don’t have enough brain power left for today to look into the implications of this. I can start with that tomorrow.

src/protocol/vm/state.rs
```diff
@@ -664,6 +642,16 @@ impl VMPoolState<PreCachedDB> {
...
+        if sell_amount_exceeds_limit {
+            return Err(TychoSimulationError::SellAmountTooHigh(
+                // Partial buy amount and gas used
+                buy_amount,
+                trade.gas_used,
+                new_state,
+                sell_amount_limit,
+            ));
+        }
         Ok((buy_amount, trade.gas_used, new_state))
```
tamaralipows, line 655: Forgot to actually translate this to a `GetAmountOutResult`. Will do it tomorrow.

src/protocol/uniswap_v3/state.rs
```diff
@@ -279,13 +280,18 @@ impl ProtocolSim for UniswapV3State {
         let result = self.swap(zero_for_one, amount_specified, None)?;
 
         trace!(?amount_in, ?token_a, ?token_b, ?zero_for_one, ?result, "V3 SWAP");
+        let mut new_state = self.clone();
+        new_state.liquidity = result.liquidity;
+        new_state.tick = result.tick;
+        new_state.sqrt_price = result.sqrt_price;
```
dianacarvalho1, line 286: I _think_ this is enough but can any uniswap v3 expert check it please?

src/protocol/vm/state.rs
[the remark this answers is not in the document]
dianacarvalho1: this is still a problem 😩

src/evm/simulation.rs
```diff
@@ -627,9 +628,9 @@ mod tests {
         let caller = Address::from_str("0x0000000000000000000000000000000000000000")?;
         let router_addr = Address::from_str("0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D")?;
         let router_abi = BaseContract::from(
-        parse_abi(&[
-            "function getAmountsOut(uint amountIn, address[] memory path) public view returns (uint[] memory amounts)",
-        ])?
+            parse_abi(&[
```
tvinagre, line 631: ok we need to have a standard for rustrover setup - I'll post it in the channel

src/protocol/dodo/state.rs
```diff
@@ -173,7 +170,11 @@ where
             .pool_abi
             .decode_output::<U256, _>("querySellBaseToken", simulation_result.result)
             .expect("DODO: Failed decoding get_amount_out result!");
-        Ok(GetAmountOutResult { amount: amount_out, gas: U256::from(simulation_result.gas_used) })
+        Ok(GetAmountOutResult {
+            amount: amount_out,
+            gas: U256::from(simulation_result.gas_used),
+            new_state: self.clone_box(), // WARNING: This is not correct but dodo is not being used
```
tvinagre, line 176: Let's remove Dodo then

tvinagre: Better than having a non-working dependency on a new package

src/protocol/uniswap_v3/state.rs
tvinagre: don't you need to update the tick list?

dianacarvalho1: I don't really know. but it isn't available in the result..

src/protocol/dodo/state.rs
dianacarvalho1: sure? I wanted to remove it some time ago but @kayibal was against it 😅

src/protocol/vm/engine.rs
```diff
@@ -48,15 +51,18 @@ where
     let db_read = db.read().await;
     let engine = SimulationEngine::new(db_read.clone(), trace);
 
+    let contract_bytecode = load_erc20_bytecode()?;
```
dianacarvalho1, line 54: my god. Thank you for finding this @tamaralipows 🙏🏼

src/protocol/vm/adapter_contract.rs
```diff
@@ -93,17 +95,48 @@ where
...
+                .first()
+                .cloned()
+                .ok_or(SimulationError::DecodingError(
+                    "Expected at least one element in the calculated price".into(),
+                ))?;
+
+            Ok((received_amount, gas_used, price))
+        } else {
+            Err(SimulationError::DecodingError("Expected return_value to be a Token::Tuple".into()))
+        }?;
```
louise-poole, line 139: This can be made easier to read/follow with pattern matching. Something like:
```suggestion
     let (received_amount, gas_used, price) = if let Token::Tuple(ref return_value) = res.return_value[0] {
        match &return_value[..] {
            [Token::Uint(amount), Token::Uint(gas), Token::Tuple(price_elements)] => {
                let received_amount = *amount;
                let gas_used = *gas;
    
                let price_token = Token::Array(vec![Token::Tuple(price_elements.clone())]);
                let price = self
                    .calculate_price(price_token)?
                    .first()
                    .cloned()
                    .ok_or_else(|| SimulationError::DecodingError(
                        "Expected at least one element in the calculated price".into(),
                    ))?;
    
                Ok((received_amount, gas_used, price))
            }
            _ => Err(SimulationError::DecodingError("Expected tuple elements with correct types".into())),
        }
    } else {
        Err(SimulationError::DecodingError("Expected return_value to be a Token::Tuple".into()))
    }?;
```

louise-poole: Also can we change the error wording to not use the 'expect' pattern?

src/protocol/vm/engine.rs
```diff
@@ -48,15 +51,18 @@ where
...
             balance: Default::default(),
             nonce: 0,
             code_hash: KECCAK_EMPTY,
-            code: None,
+            code: Some(contract_bytecode.clone()),
         };
         engine.state.init_account(
-            Address::parse_checksummed(token, None).expect("Invalid checksum for token address"),
+            Address::parse_checksummed(token, None)
+                .expect("checksum for token address must be valid"),
```
louise-poole, line 65: We should raise an error instead of panic here.

src/protocol/vm/engine.rs
```diff
@@ -180,30 +186,25 @@ mod tests {
...
             let token_address = Address::parse_checksummed(token, None).expect("valid checksum");
-            let account = engine
-                .state
-                .data
+            let account = state_data
+                .clone()
                 .borrow()
                 .get(&token_address)
                 .unwrap()
                 .clone();
```
louise-poole, line 198: I suspect we don't need to clone this often... we might be able to do something llike:
```suggestion
            let account = state_data.borrow().get(&token_address).cloned().unwrap();
```

src/protocol/vm/state.rs
```diff
@@ -541,6 +544,111 @@ impl VMPoolState<PreCachedDB> {
...
+        let mut sell_amount_exceeds_limit = false;
+        let overwrites = self
+            .get_overwrites(
+                vec![sell_token, buy_token],
+                U256::from_big_endian(&(*MAX_BALANCE / rU256::from(100)).to_be_bytes::<32>()),
+            )
+            .await?;
+
+        let sell_amount_limit = self
+            .clone()
```
louise-poole, line 564: Do we really need to clone here?

src/protocol/vm/state.rs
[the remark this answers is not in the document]
louise-poole: Looks like we no longer need to clone here now?

src/protocol/vm/state.rs
```diff
@@ -541,6 +544,111 @@ impl VMPoolState<PreCachedDB> {
...
+            .await?;
+
+        if self
+            .capabilities
+            .contains(&Capability::HardLimits) &&
+            sell_amount_limit < sell_amount
+        {
+            sell_amount_respecting_limit = sell_amount_limit;
+            sell_amount_exceeds_limit = true;
+        }
```
louise-poole, line 575:
```suggestion
        let overwrites = self
            .get_overwrites(
                vec![sell_token, buy_token],
                U256::from_big_endian(&(*MAX_BALANCE / rU256::from(100)).to_be_bytes::<32>()),
            )
            .await?;

        let sell_amount_limit = self
            .get_sell_amount_limit(vec![sell_token, buy_token], Some(overwrites.clone()))
            .await?;

        let (sell_amount_respecting_limit, sell_amount_exceeds_limit) = if self.capabilities.contains(&Capability::HardLimits) && sell_amount_limit < sell_amount {
            (sell_amount_limit, true)
        } else {
            (sell_amount, false)
        };
```

src/protocol/vm/state.rs
```diff
@@ -541,6 +544,111 @@ impl VMPoolState<PreCachedDB> {
...
+
+        let overwrites_with_sell_limit = self
+            .clone()
+            .get_overwrites(vec![sell_token, buy_token], sell_amount_limit)
+            .await?;
+        let complete_overwrites = self.merge(&overwrites, &overwrites_with_sell_limit);
+        let pool_id = self.clone().id;
```
louise-poole, line 582: Is it possible to clone only the id instead of the whole pool?
```suggestion
        let pool_id = self.id.clone();
```

src/protocol/vm/state.rs
```diff
@@ -541,6 +544,111 @@ impl VMPoolState<PreCachedDB> {
...
+
+        let (trade, state_changes) = self
+            .adapter_contract
+            .clone()
```
louise-poole, line 586: Do we need to clone here? Excessive cloning can make things very slow. We can try use `.as_ref()`

src/protocol/vm/state.rs
```diff
@@ -404,7 +414,11 @@ impl VMPoolState<PreCachedDB> {
     /// Retrieves the sell amount limit for a given pair of tokens, where the first token is treated
     /// as the sell token and the second as the buy token. The order of tokens in the input vector
     /// is significant and determines the direction of the price query.
-    async fn get_sell_amount_limit(&mut self, tokens: Vec<H160>) -> Result<U256, SimulationError> {
+    async fn get_sell_amount_limit(
+        &mut self,
```
louise-poole, line 418: This no longer needs to be mutable - so we no longer need to clone before using it in get_amount_out.

src/protocol/vm/state.rs
```diff
@@ -541,6 +544,111 @@ impl VMPoolState<PreCachedDB> {
...
+                            })?,
+                            U256::from_dec_str(&value_str).map_err(|_| {
+                                SimulationError::DecodingError(
+                                    "Failed to decode slot overwrite".to_string(),
+                                )
+                            })?,
+                        );
+                }
+            }
+        }
```
louise-poole, line 626:
```suggestion
        for (address, state_update) in state_changes {
            if let Some(storage) = state_update.storage {
                let block_overwrites = new_state.block_lasting_overwrites.entry(address).or_default();
                for (slot, value) in storage {
                    let slot = U256::from(slot).map_err(|_| SimulationError::DecodingError("Failed to decode slot index".to_string()))?;
                    let value = U256::from(value).map_err(|_| SimulationError::DecodingError("Failed to decode slot overwrite".to_string()))?;
                    block_overwrites.insert(slot, value);
                }
            }
        }
```

src/protocol/vm/state.rs
louise-poole: I'm not sure we need a whole var for sell_amount_exceeds_limit... it's only used in one place so we could just perform this check there?

src/protocol/vm/state.rs
```diff
@@ -754,25 +876,103 @@ mod tests {
...
+            .await
+            .unwrap();
+        let new_state = result
+            .new_state
+            .as_any()
+            .downcast_ref::<VMPoolState<PreCachedDB>>()
+            .unwrap();
+        assert_eq!(result.amount, U256::from_dec_str("137780051463393923").unwrap());
+        assert_eq!(result.gas, U256::from_dec_str("102770").unwrap());
+        assert_ne!(new_state.spot_prices, pool_state.spot_prices);
```
louise-poole, line 901: Can we also add checks that the block lasting overwrites are unchanged for the original pool?

src/protocol/vm/tycho_simulation_contract.rs
```diff
@@ -193,10 +194,12 @@ where
...
 #[cfg(test)]
 mod tests {
-    use super::*;
-    use revm::primitives::{hex, AccountInfo, Address, Bytecode, B256, U256 as rU256};
     use std::str::FromStr;
 
+    use revm::primitives::{hex, AccountInfo, Address, Bytecode, B256, U256 as rU256};
+
+    use super::*;
+
```
louise-poole, line 202: Hint: for these test sections if you import super first, then if you double import something that's already in super it will warn you. If you import super last it doesn't warn you.

louise-poole (commented):
Well done on figuring out the overwrites issue! Just a couple comments - most should hopefully help with test speed etc.

src/protocol/errors.rs
```diff
@@ -83,4 +84,6 @@ pub enum SimulationError {
     ArithmeticOverflow(),
     #[error("Unknown error")]
     Unknown(),
+    #[error("Sell amount is higher than sell limit")]
+    SellAmountTooHigh(), // TODO: Make it recoverable
```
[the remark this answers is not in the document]
dianacarvalho1: yes yes, this is supposed to be a partial result! It's in the list already - I just didn't want to deal with it now (particularly with the upcoming error refactor)

src/protocol/uniswap_v3/state.rs
```diff
@@ -152,6 +152,7 @@ impl UniswapV3State {
                         return Err(SimulationError::InsufficientData(GetAmountOutResult::new(
                             state.amount_calculated.abs().into_raw(),
                             gas_used,
+                            self.clone_box(),
```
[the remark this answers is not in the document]
dianacarvalho1: huum do you think that is a good behaviour though? like if it changed, we pass it, if it didn't change we don't pass it? 🤔 I'm not sure what would be best for the user tbh. they would have to check if it is some and then use the same state or the new one

src/protocol/vm/state.rs
dianacarvalho1: true, now that we removed the mutable thingies I think we can remove it 👀

src/protocol/uniswap_v3/state.rs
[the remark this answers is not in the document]
louise-poole: Performance wise it is much more efficient to check for a None than to always clone.
However, looking closer at this we should be returning a new state here no? We are providing a partial result so should return a new state as if that partial swap had been completed.

src/protocol/uniswap_v3/state.rs
louise-poole: Update - it's not needed. The tick list does not change between swaps.

PR #36, opened 2024-11-05
chore: Make price printer an example

kayibal:
Running the example is now very simple:

```
RUST_LOG=info cargo run --example price-printer
```


louise-poole (approved):
Just fix CI and AG 👍

PR #40, opened 2024-11-05
feat: Refactor Error handling

tamaralipows:
[More info](https://propellerswap.slack.com/archives/C05LQALE9L1/p1730799820560479)


src/protocol/uniswap_v2/state.rs
```diff
@@ -93,22 +94,22 @@ impl ProtocolSim for UniswapV2State {
         amount_in: U256,
         token_in: &ERC20Token,
         token_out: &ERC20Token,
-    ) -> Result<GetAmountOutResult, TradeSimulationError> {
+    ) -> Result<GetAmountOutResult, TychoSimulationError> {
         if amount_in == U256::zero() {
-            return Result::Err(TradeSimulationError::new(
+            return Result::Err(TychoSimulationError::from(NativeSimulationError::new(
```
dianacarvalho1, line 99: oh I would expect that we don't have to explicitly convert this like this 😬 
(I checked locally and I see we actually do buuuuuh)

dianacarvalho1 (approved):
I don't understand why it shows that the cargo lock was deleted when @tamaralipows added it! I think github is cray cray. Thank you for taking this! I'm gonna merge it and rebase the other branch on it! 🙏🏼

dianacarvalho1:
there is still something else to do here: return a result from the spot prices method (I will do that next)

src/protocol/errors.rs
```diff
@@ -49,3 +96,27 @@ pub enum InvalidSnapshotError {
...
+///
+/// `TychoSimulationError` encompasses all possible errors that can occur in the package,
+/// wrapping lower-level errors in a user-friendly way for easier handling and display.
+#[derive(Error, Debug)]
+pub enum TychoSimulationError {
+    #[error("VM simulation error: {0}")]
+    VMError(VMError),
+    #[error("Native simulation error: {0}")]
+    NativeSimulationError(NativeSimulationError),
+}
```
kayibal, line 110: Why are we differentiating between these errors (vm vs native) - I don't think we should make a differentiation here as this is an implementation specific detail that shouldn't leak into the interface! 

I think we should merge all of the cases this into a single error struct. Let's also get rid of the `...ErrorKind` enum.

src/protocol/errors.rs
```diff
@@ -18,19 +19,65 @@ pub enum TradeSimulationErrorKind {
...
+            }
+            TradeSimulationErrorKind::InsufficientAmount => {
+                write!(f, "Insufficient amount provided for the trade")
+            }
+            TradeSimulationErrorKind::U256Overflow => {
+                write!(f, "Arithmetic operation resulted in a U256 overflow")
+            }
+        }
+    }
+}
```
kayibal, line 60: I understand we had to implement Display here because we are using this weird pattern of having an `ErrorKind` instead of just having a single enum for all errors. I much prefer the usage of the `#[error(...)]` macro over this.

kayibal (changes requested):

src/protocol/dodo/state.rs
```diff
@@ -145,7 +145,7 @@ where
         _token_out: &crate::models::ERC20Token,
     ) -> Result<
         crate::protocol::models::GetAmountOutResult,
-        crate::protocol::errors::TradeSimulationError,
+        crate::protocol::errors::TychoSimulationError,
```
tvinagre, line 148: I might have missed a discussion but is `Tycho` necessary here?

src/protocol/vm/state.rs
```diff
@@ -203,7 +204,9 @@ impl VMPoolState<PreCachedDB> {
                 } else {
                     let code =
                         Bytecode::new_raw(Bytes::from(bytecode.clone().ok_or_else(|| {
-                            TychoSimulationError::DecodingError("Expected bytecode".into())
+                            TychoSimulationError::from(VMError::DecodingError(
+                                "Expected bytecode".into(),
```
louise-poole, line 208: Usually when we use a `.expect` we phrase the error message as what was expected. Here when we raise an error we phrase the message as what went wrong. In this case it looks like we're raising because the bytecode is None?

src/protocol/vm/state.rs
```diff
@@ -279,25 +286,34 @@ impl VMPoolState<PreCachedDB> {
...
+        let address = decode(&[ParamType::Address], &sim_result.result)
+            .map_err(|_| {
+                TychoSimulationError::from(VMError::DecodingError("Failed to decode ABI".into()))
+            })?
             .into_iter()
             .next()
             .ok_or_else(|| {
-                TychoSimulationError::DecodingError("Expected an address in the result".into())
+                TychoSimulationError::from(VMError::DecodingError(
+                    "Expected an address in the result".into(),
```
louise-poole, line 299: Same with these error messages with the expect wording. To an outside user this will make no sense... the engine raises an error that some sort of internal result had no address in it? It's not clear to me how to fix that or what actually went wrong.

src/protocol/vm/state.rs
```diff
@@ -279,25 +286,34 @@ impl VMPoolState<PreCachedDB> {
...
+                ))
             })?;
 
         address
             .to_string()
             .parse()
-            .map_err(|_| TychoSimulationError::DecodingError("Expected an Address".into()))
+            .map_err(|_| {
+                TychoSimulationError::from(VMError::DecodingError("Expected an Address".into()))
```
louise-poole, line 307: This error isn't a missing address, it's a badly encoded one. This will be more like `failed to decode` than 'expected a missing address'

src/protocol/vm/state.rs
```diff
@@ -385,11 +401,11 @@ impl VMPoolState<PreCachedDB> {
                 .contains(&Capability::ScaledPrice)
             {
                 *price_result.first().ok_or_else(|| {
-                    TychoSimulationError::DecodingError("Expected a u64".to_string())
+                    TychoSimulationError::from(VMError::DecodingError("Expected a u64".to_string()))
```
louise-poole, line 404: Here too. We need more details to help someone figure out why this error is occuring. What was meant to be a u64?

src/protocol/vm/state.rs
```diff
@@ -502,7 +518,9 @@ impl VMPoolState<PreCachedDB> {
         let address = match self.balance_owner {
             Some(address) => Ok(address),
             None => self.id.parse().map_err(|_| {
-                TychoSimulationError::EncodingError("Pool ID is not an address".into())
+                TychoSimulationError::from(VMError::EncodingError(
+                    "Pool ID is not an address".into(),
```
louise-poole, line 522: Yay - perfection.

louise-poole (commented):
Just a couple comments 👀  A few of them could be moved to the clean up task if we want to not do them right now.

src/protocol/errors.rs
```diff
@@ -1,4 +1,11 @@
 //! Protocol generic errors
+use crate::{
+    evm::simulation::SimulationError,
+    protocol::{
+        errors::TychoSimulationError::VMError,
+        vm::errors::{FileError, RpcError, VMError},
+    },
+};
```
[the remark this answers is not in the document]
dianacarvalho1: it's fixed in a later commit ✨

src/protocol/vm/errors.rs
```diff
@@ -6,46 +6,7 @@ use ethers::prelude::ProviderError;
...
-    #[error("Decoding error: {0}")]
-    DecodingError(String),
-    #[error("RPC related error {0}")]
-    RpcError(RpcError),
-    #[error("Unsupported Capability: {0}")]
-    UnsupportedCapability(String),
-    #[error("Adapter not initialized: {0}")]
-    UninitializedAdapter(String),
-    #[error("Engine not set")]
-    EngineNotSet(),
```
kayibal, line 47: Pretty much all of these except `SimulationFailure` seem to be Fatal so you can map them to `Unkown()`

kayibal (commented):
A note on the reason of this change: it is not a fix, we change a lot of interfaces here. Since we are at 0.x version this means we should only increment the minor version for now so let's please make this at least `feat:`.

src/protocol/errors.rs
```diff
@@ -49,3 +28,51 @@ pub enum InvalidSnapshotError {
...
+/// - `InsufficientData`: Error indicating that there is insufficient data to perform the
+///   simulation. It returns a partial result of the simulation.
+/// - `NoLiquidity`: Error indicating that there is no liquidity in the venue to complete the trade.
+/// - `InsufficientAmount`: Error indicating that the amount provided for the trade is insufficient.
+/// - `U256Overflow`: Error indicating that an arithmetic operation got an U256 to overflow
+/// - `Unknown`: Error indicating that an unknown error occurred during the simulation.
+#[derive(Error, Debug)]
+pub enum SimulationError {
+    #[error("ABI loading error: {0}")]
+    AbiError(FileError),
```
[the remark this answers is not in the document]
dianacarvalho1: aahh that is enough?? we are implementing the from separately! will change it

src/protocol/vm/adapter_contract.rs
```diff
@@ -3,9 +3,12 @@
 
 use crate::{
     evm::account_storage::StateUpdate,
-    protocol::vm::{
-        erc20_overwrite_factory::Overwrites, errors::TychoSimulationError, models::Capability,
-        tycho_simulation_contract::TychoSimulationContract,
+    protocol::{
+        errors::SimulationError,
```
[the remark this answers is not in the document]
dianacarvalho1: this has gone back and forth multiple times.. 😩 now you are suggesting to separate this again. I think this is more a form concern than functionality at this point. I understand your points but I don't think this is crucial to have for devcon. I would like to merge this PR as is so we can actually test this package and make sure it works. We will come back later to this and polish things out when we have more time.

kayibal (approved):

PR #45, opened 2024-11-06
feat(simulation-py): make token bruteforce compatible with vyper

zizou0x:
This PR aims to make our token overwrite factory properly handle contract compiled with vyper (Curve).


tycho_simulation_py/python/tycho_simulation_py/evm/pool_state.py
```diff
@@ -97,8 +98,8 @@ def __init__(
         self.involved_contracts: set[Address] = involved_contracts or set()
         """A set of all contract addresses involved in the simulation of this pool."""
 
-        self.token_storage_slots: dict[Address, tuple[int, int]] = (
-            token_storage_slots or {}
+        self.token_storage_slots: dict[Address, tuple[tuple[int, int], ContractCompiler]] = (
```
kayibal, line 101: Can you mention a bit why we have this here per contract in the docs? (I think it makes a lot of sense)

tycho_simulation_py/python/tycho_simulation_py/evm/utils.py
```diff
@@ -194,6 +200,8 @@ def get_storage_slot_at_key(key: Address, mapping_slot: int) -> int:
     key_bytes = bytes.fromhex(key[2:]).rjust(32, b"\0")
     mapping_slot_bytes = int.to_bytes(mapping_slot, 32, "big")
     slot_bytes = eth_utils.keccak(key_bytes + mapping_slot_bytes)
+    if lang == ContractCompiler.Vyper:
+        slot_bytes = eth_utils.keccak(mapping_slot_bytes + key_bytes)
```
kayibal, line 204: Would it make sense to move this logic directly to `ContractCompiler`? Could make things a bit more obvious but it's not necessary imo.

kayibal (approved):
Just some nitpicks

PR #46, opened 2024-11-06
feat: (WIP) move `get_amount_out` into `ProtocolSim` trait

tamaralipows:
- In order to do this, we need to make the trait async
- Can't parametrize using rstest on an async fn
- Unfortunately this breaks the UI because `terminal.draw(|frame| self.draw(frame).await)?;` doesnt accept an async fn


examples/explorer/ui.rs
```diff
@@ -182,7 +182,7 @@ impl App {
     pub async fn run(mut self, mut terminal: DefaultTerminal) -> anyhow::Result<()> {
         let mut reader = event::EventStream::new();
         loop {
-            terminal.draw(|frame| self.draw(frame))?;
+            terminal.draw(|frame| self.draw(frame).await)?;
```
tamaralipows, line 185: problem

PR #47, opened 2024-11-07
fix: return new state on partially successful simulations

louise-poole:
Issue: we were cloning the original state and returning that with the partial trade. This is inconsistent with all other uses of `GetAmountOutResult` where the state returned is a new one after the trade.

Solution: return a new_state that is updated to represent the state after the partial trade.


dianacarvalho1 (approved):
Thank you @louise-poole 🙏🏼

PR #50, opened 2024-11-07
feat: Add balancer to example

tamaralipows:
Review by commit


Cargo.toml
```diff
@@ -28,8 +28,8 @@ uuid = { version = "1.4.1", features = [
 ] }
 num-traits = "0.2.17"
 dotenv = "0.15.0"
-tycho-core = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-core", tag = "0.41.0" }
-tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-client", tag = "0.41.0" }
+tycho-core = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-core", rev = "baf2619d8021b6ebcfe06a04774249c79aba25e9"  }
+tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-client", rev = "baf2619d8021b6ebcfe06a04774249c79aba25e9"  }
```
tamaralipows, line 32: This can be changed back once we merge the pr in the tycho indexer repo

kayibal: I think it would be better to do things the other way around, ntq should use tycho-simulation version. We could think about re-exporting tycho-core on the tycho-simulation crate to avoid accidentally using two different versions in dependencies: https://doc.rust-lang.org/rustdoc/write-documentation/re-exports.html

kayibal (approved):

examples/explorer/data_feed/tycho.rs
```diff
@@ -137,29 +234,63 @@ pub async fn process_messages(
...
-
+            let storage_by_address: HashMap<Address, ResponseAccount> = protocol_msg
+                .clone()
+                .snapshots
+                .get_vm_storage()
+                .into_iter()
+                .map(|(key, value)| (Address::from_slice(&key[..20]), value.clone().into()))
+                .collect();
+            info!("Updating engine with snapshot");
+            update_engine(SHARED_TYCHO_DB.clone(), block, Some(storage_by_address), HashMap::new())
```
dianacarvalho1, line 291: This needs to be done before creating the components. The vm storage needs to be in the engine before we actually make the pools. we are setting spot prices on creation now so we need to simulate right away. For balancer case, not updating the engine first is making the balancer vault unavailable and the simulations fail miserably

tamaralipows:
Merging this now so that it doesn't block my cleanup tasks. Feel free to review anyway. Fixes will be made in future (smaller) PRs.

PR #51, opened 2024-11-08
feat(simulation): add bruteforce token slots detection

zizou0x:
Add a function that will try to brute force balance and allowance maps base slots for a given token. 

This function will iterate over the mapping "base_slot" and try to set a marker value to a particular key and read it. If the value is correctly retrieved, it means the "base_slot" is correct.

Also add a support for Vyper contracts for which the storage layout is different.


src/evm/mod.rs
```diff
@@ -1,7 +1,48 @@
...
 pub mod engine_db_interface;
 pub mod simulation;
 pub mod simulation_db;
+pub mod token;
 pub mod traces;
 pub mod tycho_db;
 pub mod tycho_models;
+
+/// Enum representing the type of contract compiler.
+#[derive(Debug, PartialEq, Copy, Clone)]
```
zizou0x, line 15: I made it `Copy` because it's a cheap enum

src/evm/token.rs
```diff
@@ -0,0 +1,248 @@
...
+    static ref ERC20_ABI: Abi = {
+        let abi_file_path = "src/protocol/vm/assets/ERC20.abi";
+        let abi_json = std::fs::read_to_string(abi_file_path).expect("Failed to read ABI file");
+        from_str(&abi_json).expect("Failed to parse ABI JSON")
+    };
+}
+
+/// An error r
+#[derive(Error, Debug)]
+pub enum TokenError {
```
zizou0x, line 29: Is it overkill? I can make it a `String` but I like to have specific types for errors. Or if you have any suggestion about where this error could be added?

tamaralipows: I think this has to be a `SimulationError`

src/protocol/vm/state.rs
```diff
@@ -522,6 +536,24 @@ impl VMPoolState<PreCachedDB> {
...
+                self.token_storage_slots.insert(
+                    *t,
+                    token::brute_force_slots(
+                        t,
+                        &self.block,
+                        self.engine
+                            .as_ref()
+                            .expect("engine should be set"),
+                    )
+                    .unwrap_or_else(|_| panic!("Couldn't brute force slots for {}", t)),
```
tamaralipows, line 552: I think we should return the error here like we do in the rest of the file.

tamaralipows (approved):
Thank you! Looks great

PR #52, opened 2024-11-08
fix: fix protocol to adapter file path conversion

louise-poole:
and update docs and remove dead code clippy skips.


dianacarvalho1 (commented):
Thank you 🙏🏼

Readme.md
```diff
@@ -13,26 +13,31 @@ protocols, but it is not necessarily limited to this.
...
 
 ## Adding a new Protocol
 
 To add a new protocol, you will need to complete the following high-level steps:
 
+#### Native protocol:
+
 1. Create a protocol state struct that implements the `ProtocolSim` struct.
-2. _(legacy information)_ Add associated events and implement the transition method on the protocol state struct.
+2. Create a tycho decoder for the protocol state: i.e. implement `TryFrom` for `ComponentWithState` to your new
```
kayibal, line 31: Is it `TryFrom` or `TryFromWithBlock`, also we should specify the error type of the implementation here I think.

src/protocol/vm/tycho_decoder.rs
```diff
@@ -113,15 +117,15 @@ impl TryFromWithBlock<ComponentWithState> for VMPoolState<PreCachedDB> {
...
-                    .protocol_system
-                    .as_str(),
-            )
-        );
+        let protocol_name = snapshot
+            .component
+            .protocol_system
+            .strip_prefix("vm:")
+            .unwrap_or({
+                Err(InvalidSnapshotError::VMError(SimulationError::DecodingError("Invalid protocol system name: VMPoolStates expect a protocol name starting with 'vm:'".to_string())))?
```
kayibal: What err inception is going on here? Did you just look for the closest DecodingError?

src/protocol/vm/tycho_decoder.rs
```diff
@@ -113,15 +117,15 @@ impl TryFromWithBlock<ComponentWithState> for VMPoolState<PreCachedDB> {
...
-                snapshot
-                    .component
-                    .protocol_system
-                    .as_str(),
-            )
-        );
+        let protocol_name = snapshot
+            .component
+            .protocol_system
+            .strip_prefix("vm:")
```
kayibal, line 123: What if we have a VM and a native implementation?

Readme.md
louise-poole: It's `TryFrom`. `TryFromWithBlock is only used by VmPoolState which already has a decoder.

src/protocol/vm/tycho_decoder.rs
louise-poole: `InvalidSnapshotError` only has 'MissingAttribute', 'VMError' and 'ValueError'. Initially felt VMError was the most appropriate, but a simple ValueError is probably better?

src/protocol/vm/tycho_decoder.rs
louise-poole: True! I'll make the stripping optional.

louise-poole: On that note, is there a reason we still need the 'vm:' prefix for these protocol names in tycho?

dianacarvalho1 (approved):

Readme.md
kayibal: Would prefer if we were consistent

src/protocol/vm/tycho_decoder.rs
kayibal: I guess if we have both a VM and native implementation we could distinguish them like this.

Readme.md
louise-poole: Okay, I'll do this in a separate PR... Diana is waiting on this one.

PR #55, opened 2024-11-10
refactor: use TryFromWithState for native decoders too

louise-poole:


src/protocol/models.rs
```diff
@@ -50,6 +54,19 @@ impl ProtocolComponent {
...
+
+    #[allow(async_fn_in_trait)]
+    async fn try_from_with_block(
+        value: T,
+        block: Header,
+        all_tokens: HashMap<H160, ERC20Token>,
+    ) -> Result<Self, Self::Error>
+    where
+        Self: Sized;
+}
```
dianacarvalho1, line 68: I have a generic rust question:
Is it normal to name the trait `DoSomething` and then the method `do_something` as well? isn't it redundant? or this is just rust rusting? I'm asking because of `TryFromWithBlock` and `try_from_with_block`

dianacarvalho1 (approved):
Thank you so much @louise-poole 🙏🏼 I only have a question to increase my rust knowledge 👀 Otherwise it looks good to me!

PR #57, opened 2024-11-15
chore: Make EngineDatabaseInterface rely on DatabaseRef

tamaralipows:
Small simplification. Like this, EngineDatabaseInterface can only be implemented on those which implement the DatabaseRef trait, which makes sense for our case.

### DIFF
--- a/src/evm/engine_db_interface.rs
+++ b/src/evm/engine_db_interface.rs
@@ -3,9 +3,10 @@ use std::collections::HashMap;
 use revm::{
     precompile::Address,
     primitives::{AccountInfo, U256 as rU256},
+    DatabaseRef,
 };
 
-pub trait EngineDatabaseInterface {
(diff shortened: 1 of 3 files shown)


zizou0x (approved):

src/evm/engine_db_interface.rs
```diff
@@ -3,9 +3,10 @@ use std::collections::HashMap;
 use revm::{
     precompile::Address,
     primitives::{AccountInfo, U256 as rU256},
+    DatabaseRef,
 };
 
-pub trait EngineDatabaseInterface {
+pub trait EngineDatabaseInterface: DatabaseRef {
```
dianacarvalho1, line 9: oh uau! I didn't know about this subtrait thingy! Thank you @tamaralipows 🙏🏼

PR #58, opened 2024-11-19
feat: Implement clear_all_cache

tamaralipows:
- Biggest thing that was missing was clearing the temp storage of the simulation engine. The spot prices were already being re-set.


src/protocol/vm/state.rs
```diff
@@ -422,6 +427,16 @@ impl VMPoolState<PreCachedDB> {
         Ok(limits?.0)
     }
 
+    fn clear_all_cache(&mut self, tokens: Vec<ERC20Token>) -> Result<(), SimulationError> {
+        self.clone()
```
zizou0x, line 431: Why clone? We're have a `&mut` we should be able to clear the storage from it no?

zizou0x (approved):

PR #59, opened 2024-11-19
fix: avoid deadlocks in `PrecachedDB`

zizou0x:
This PR fixes an issue that we were getting when running the example. There was a deadlock in the PrecachedDB that was causing the program to hang indefinitely. 

This PR fixes it by using a sync RwLock to avoid this kind or unexpected behavior. Also it helps to simplify the code by removing the need of the `block_on` function.

This PR also fixes the `trace` feature for simulation which was broken because of a tokio runtime blocking issue.


tamaralipows (approved):
Thank you!
Repeated from slack:

I think the reason we used tokio instead of std was because the tokio RwLock prioritizes writes. From the [story](https://datarevenue.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-3727):
Also we need a mechanism to update this database, meaning that if an update comes in, it eventually is allowed to write to the DB even though many threads have a read lock on it and the update is atomic, meaning that data doesn’t change while some pool is simulating a transaction.
Concerning this I’m not sure if there are any negative implications of your changes… but other than this from what I see it looks great and simpler! Perhaps we can merge it, keep in mind this issue, and if someone with more experience has any objections in the future we can revert or fix in a separate PR?

src/evm/tycho_db.rs
```diff
@@ -156,19 +147,14 @@ impl PreCachedDB {
     ///
     /// Returns an `Option` containing a reference to the storage value if it exists, otherwise
     /// returns `None`.
-    async fn get_storage_async(&self, address: &Address, index: &rU256) -> Option<rU256> {
+    pub fn get_storage(&self, address: &Address, index: &rU256) -> Option<rU256> {
         self.inner
             .read()
-            .await
+            .unwrap()
```
dianacarvalho1, line 153: can you explain why it is ok to use the `.unwrap()` here?? asking just to understand rust a bit better

PR #61, opened 2024-11-19
refactor: clean up account inits

tamaralipows:
[This task](https://datarevenue.atlassian.net/browse/ENG-3829) had three questions:

**Does it make sense to init the adapter contract in the pool state and not in the adapter contract?**
- No. I changed this (first commit).
- This way we can isolate adapter-contract-related functionality within the adapter contract and this information is not leaked inside the pool state.

**Does it make sense that we are initing accounts in create_engine and in set_engine?**
- No this was messy. It has now been moved entirely to `set_engine`.
- It may have been nice to also move this inside create_engine, but, since we need the `get_address_from_call` method, the refactoring required for that was extensive.
- Moving `get_address_from_call` into utils required it to be passed the engine with a generic parameter instead of `PreCachedDB` since it would be called in `create_engine`. This required in a huge chain of required changes due to most of our classes, including the simulation, not deriving the `DEBUG` trait. I wasn't convinced this change was worth it.
- It also makes sense for the `create_engine` method to do exactly only that - create the engine. No hidden account initializations.
- The test in engine.rs was highly, highly mocked. By moving inits into `set_engine`, we are able to truly test that the account inside PreCachedDB is successfully initialized. This makes our tests more reliable.

**Does it make sense to pass the engine around like we are doing now?**
- I've tried to eliminate the clones using lifetimes, and, when that failed, using Arc paired with Mutex. For each of these techniques, it was a massive massive change, and even after fixing many many files to get it to compile, I was still getting compilation errors. I decided to abort until we are sure we want to dedicate more time to this.


src/protocol/vm/state.rs
```diff
@@ -139,52 +135,63 @@ impl VMPoolState<PreCachedDB> {
...
-                rAddress::parse_checksummed(ADAPTER_ADDRESS.to_string(), None)
+                rAddress::parse_checksummed("0x0000000000000000000000000000000000000004", None)
                     .expect("Invalid checksum for external account address"),
-                AccountInfo {
-                    balance: *MAX_BALANCE,
-                    nonce: 0,
-                    code_hash: B256::from(keccak256(adapter_contract_code.clone().bytes())),
-                    code: Some(adapter_contract_code),
-                },
+                zero_account_info.clone(),
```
zizou0x, line 194: Maybe we can do this in the engine creation logic? These accounts (0 and 4) are created to enable precompiles in the vm. I think we can create them by default in the engine creation?

src/protocol/vm/state.rs
```diff
@@ -139,52 +135,63 @@ impl VMPoolState<PreCachedDB> {
         Ok(state)
     }
 
-    async fn set_engine(&mut self, adapter_contract_path: String) -> Result<(), SimulationError> {
+    async fn set_engine(&mut self) -> Result<(), SimulationError> {
```
zizou0x, line 138: I still don't quite like this function though... I expect a `set_xxx` function to have a `xxx` as arg and set it on the struct. Here we are actually ensuring that the engine is set and creating one if not. I think something like `ensure_engine` would be more accurate. What do you think?

Also I think it would be nice to have a way to actually pass the engine, (that would help to fix slow tests) but probably out of scope for this PR

src/protocol/vm/state.rs
```diff
@@ -209,7 +216,8 @@ impl VMPoolState<PreCachedDB> {
                     (Some(code.clone()), code.hash_slow())
                 };
                 engine.state.init_account(
-                    address.parse().unwrap(),
+                    rAddress::parse_checksummed(address, None)
```
zizou0x, line 219: Is this needed? Also have you tested it? AFAIK addresses are given full lower cap by Tycho, so I'm afraid this will complain about addresses not being checksummed

src/evm/account_storage.rs
```diff
@@ -27,7 +27,7 @@ pub struct StateUpdate {
 #[derive(Clone, Default, Debug)]
 /// A simpler implementation of CacheDB that can't query a node. It just stores data.
 pub struct AccountStorage {
-    accounts: HashMap<Address, Account>,
+    pub accounts: HashMap<Address, Account>,
```
zizou0x, line 30: Is this change done for testing purpose only?

src/protocol/vm/state.rs
```diff
@@ -875,6 +879,38 @@ mod tests {
...
+            .engine
+            .unwrap()
+            .state
+            .inner
+            .read()
+            .await
+            .clone()
+            .accounts
+            .accounts;
+        for token in pool_state.tokens.clone() {
```
zizou0x, line 894: As far as I can tell this is the only reason why you exposed the `AccountState`? If so I'd say we create a getter function for it that would be `#[cfg(test)]`

AFAIK it's good practice to keep things as private as possible to avoid people doing unexpected things

zizou0x (approved):
Looks good! Just a few small suggestions

src/protocol/vm/state.rs
tamaralipows: Hmmm these `account_inits` were originally scattered between the two so we thought it would be better organization to keep them in one place. It's a good argument though - we could instead divide them between the tokens/stateless accounts (which are dynamically determined) in `set_engine`, and the 0 and 4 which should never change and are necessary for precompiles in `create_engine`... if you don't think that's too messy (I'll experiment now)

src/protocol/vm/state.rs
tamaralipows: great idea - thank you!

src/protocol/vm/state.rs
tamaralipows: Merging without this since I understand you will take care of it in your PR!

src/protocol/vm/engine.rs
```diff
@@ -34,6 +38,26 @@ where
     <D as DatabaseRef>::Error: Debug,
 {
     let engine = SimulationEngine::new(db.clone(), trace);
+
+    let zero_account_info =
```
dianacarvalho1, line 42: @tamaralipows I actually like this!! this way we don't need to re init all of these every time we make a new pool state!

dianacarvalho1:
Thank you for this @tamaralipows 🙏🏼 This makes sense and I'm happy we cleaned this up!

PR #62, opened 2024-11-21
refactor: Pass pool id as vec to adapter

tamaralipows:
- This makes it match the format of the other input args closer - and we now call Token::FixedBytes similar to how we do with the other arguments.
- Also move the byte slicing into `hexstring_to_vec` so we don't need to look at such ugly things inside the pool state class.


zizou0x (approved):
LGTM thanks!

dianacarvalho1 (commented):
aafff soo much better!! 🙏🏼 thank you!

PR #66, opened 2024-11-22
feat: (WIP/Sketch) add TVL to ProtocolSim

tamaralipows:
Questions:
- We actually have the TVL inside Tycho - I'm assuming - should we maybe expose that inside of `ComponentWithState` instead of re-deriving it from the balances?
- Should we give TVL in ETH? Or the quote token? I would assume in ETH. Where then, do we get the ETH price? For now, I've started designing this using the quote token...

### DIFF
--- a/examples/explorer/main.rs
+++ b/examples/explorer/main.rs
@@ -7,7 +7,7 @@ use clap::Parser;
 use data_feed::{state::BlockState, tycho};
 use futures::future::select_all;
 use std::env;
-use tokio::{sync::mpsc, task::JoinHandle};
+use tokio::{sync::mpsc, task, task::JoinHandle};
 use tracing_subscriber::{fmt, EnvFilter};
 
 #[derive(Parser)]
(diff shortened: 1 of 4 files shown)


examples/explorer/main.rs
```diff
@@ -36,32 +36,21 @@ async fn main() {
...
+    let consumer_handle = task::spawn(async move {
+        while let Some(message) = tick_rx.recv().await {
+            // println!("got message: {:?}", message);
+        }
+        Ok(())
     });
-    let tasks = [tycho_message_processor, terminal_app];
+    let tasks = [tycho_message_processor, consumer_handle];
+
     let _ = select_all(tasks).await;
```
tamaralipows, line 55: Accidental commit. Revert later.

kayibal:
Do we really need this interface change? What is the use case of having tvl exposed on the main interface?

dianacarvalho1:
> Do we really need this interface change? What is the use case of having tvl exposed on the main interface?

This was brought up by Markus during Edge City. That his first application would be to filter/order the pools by TVL..

tamaralipows:
Closing this for two reasons:
1.  It's not in the docs so it seems very non-essential.
2.  It does not seem like an easy thing to do. It also requires an interface change - so it's not a small issue.

PR #67, opened 2024-11-26
feat: Define better user-facing errors

tamaralipows:
Please refer to commit messages for more details/reasoning.

### DIFF
--- a/src/protocol/uniswap_v3/state.rs
+++ b/src/protocol/uniswap_v3/state.rs
@@ -152,11 +152,14 @@ impl UniswapV3State {
                         new_state.liquidity = state.liquidity;
                         new_state.tick = state.tick;
                         new_state.sqrt_price = state.sqrt_price;
-                        return Err(SimulationError::InsufficientData(GetAmountOutResult::new(
-                            state.amount_calculated.abs().into_raw(),
-                            gas_used,
-                            Box::new(new_state),
-                        )));
+                        return Err(SimulationError::RetryDifferentInput(
+                            "Insufficient data".into(),
+                            Some(GetAmountOutResult::new(
+                                state.amount_calculated.abs().into_raw(),
+                                gas_used,
+                                Box::new(new_state),
+                            )),
+                        ));
                     }
                     _ => return Err(SimulationError::FatalError("Unknown error".to_string())),
                 },
@@ -626,15 +629,20 @@ mod tests {
             .unwrap_err();
 
         match err {
-            SimulationError::InsufficientData(ref e) => {
-                assert_eq!(e.amount, exp);
-                let new_state = e
-                    .new_state
-                    .as_any()
-                    .downcast_ref::<UniswapV3State>()
-                    .unwrap();
-                assert_ne!(new_state.tick, pool.tick);
-                assert_ne!(new_state.liquidity, pool.liquidity);
+            SimulationError::RetryDifferentInput(ref _err, ref amount_out_result) => {
+                match amount_out_result {
+                    Some(amount_out_result) => {
+                        assert_eq!(amount_out_result.amount, exp);
+                        let new_state = amount_out_result
+                            .new_state
+                            .as_any()
+                            .downcast_ref::<UniswapV3State>()
+                            .unwrap();
+                        assert_ne!(new_state.tick, pool.tick);
+                        assert_ne!(new_state.liquidity, pool.liquidity);
+                    }
+                    _ => panic!("Partial amount out result is None. Expected partial result."),
+                }
             }
             _ => panic!("Test failed: was expecting a SimulationError::InsufficientData"),
         }
--- a/src/protocol/uniswap_v3/state.rs
+++ b/src/protocol/uniswap_v3/state.rs
@@ -111,7 +111,7 @@ impl UniswapV3State {
         sqrt_price_limit: Option<U256>,
     ) -> Result<SwapResults, SimulationError> {
         if self.liquidity == 0 {
-            return Err(SimulationError::NoLiquidity());
+            return Err(SimulationError::RetryLater("No liquidity".to_string()));
         }
         let price_limit = if let Some(limit) = sqrt_price_limit {
             limit
@@ -158,7 +158,7 @@ impl UniswapV3State {
                             Box::new(new_state),
                         )));
                     }
-                    _ => return Err(SimulationError::Unknown()),
+                    _ => return Err(SimulationError::FatalError("Unknown error".to_string())),
                 },
             };
 
--- a/src/protocol/uniswap_v3/state.rs
+++ b/src/protocol/uniswap_v3/state.rs
@@ -153,7 +153,7 @@ impl UniswapV3State {
                         new_state.tick = state.tick;
                         new_state.sqrt_price = state.sqrt_price;
                         return Err(SimulationError::RetryDifferentInput(
-                            "Insufficient data".into(),
+                            "Ticks exceeded".into(),
                             Some(GetAmountOutResult::new(
                                 state.amount_calculated.abs().into_raw(),
                                 gas_used,
--- a/src/protocol/uniswap_v3/state.rs
+++ b/src/protocol/uniswap_v3/state.rs
@@ -111,7 +111,7 @@ impl UniswapV3State {
         sqrt_price_limit: Option<U256>,
     ) -> Result<SwapResults, SimulationError> {
         if self.liquidity == 0 {
-            return Err(SimulationError::RetryLater("No liquidity".to_string()));
+            return Err(SimulationError::RecoverableError("No liquidity".to_string()));
         }
         let price_limit = if let Some(limit) = sqrt_price_limit {
             limit
@@ -152,7 +152,7 @@ impl UniswapV3State {
                         new_state.liquidity = state.liquidity;
                         new_state.tick = state.tick;
                         new_state.sqrt_price = state.sqrt_price;
-                        return Err(SimulationError::RetryDifferentInput(
+                        return Err(SimulationError::InvalidInput(
                             "Ticks exceeded".into(),
                             Some(GetAmountOutResult::new(
                                 state.amount_calculated.abs().into_raw(),
@@ -629,7 +629,7 @@ mod tests {
             .unwrap_err();
 
         match err {
-            SimulationError::RetryDifferentInput(ref _err, ref amount_out_result) => {
+            SimulationError::InvalidInput(ref _err, ref amount_out_result) => {
                 match amount_out_result {
                     Some(amount_out_result) => {
                         assert_eq!(amount_out_result.amount, exp);
--- a/src/protocol/vm/adapter_contract.rs
+++ b/src/protocol/vm/adapter_contract.rs
@@ -91,30 +91,38 @@ where
 
         let res = self.call("swap", args, block, None, overwrites, None, U256::zero())?;
 
-        let (received_amount, gas_used, price) = if let Token::Tuple(ref return_value) =
-            res.return_value[0]
-        {
-            match &return_value[..] {
-                [Token::Uint(amount), Token::Uint(gas), Token::Tuple(price_elements)] => {
-                    let received_amount = *amount;
-                    let gas_used = *gas;
+        let (received_amount, gas_used, price) =
+            if let Token::Tuple(ref return_value) = res.return_value[0] {
+                match &return_value[..] {
+                    [Token::Uint(amount), Token::Uint(gas), Token::Tuple(price_elements)] => {
+                        let received_amount = *amount;
+                        let gas_used = *gas;
 
-                    let price_token = Token::Array(vec![Token::Tuple(price_elements.clone())]);
-                    let price = self
-                        .calculate_price(price_token)?
-                        .first()
-                        .cloned()
-                        .ok_or_else(|| {
-                            SimulationError::FatalError("There wasn't a calculated price".into())
-                        })?;
+                        let price_token = Token::Array(vec![Token::Tuple(price_elements.clone())]);
+                        let price =
+                            self.calculate_price(price_token)?
+                                .first()
+                                .cloned()
+                                .ok_or_else(|| {
+                                    SimulationError::FatalError(
+                                "Adapter swap call failed: An empty price list was returned".into(),
+                            )
+                                })?;
 
-                    Ok((received_amount, gas_used, price))
+                        Ok((received_amount, gas_used, price))
+                    }
+                    _ => Err(SimulationError::FatalError(format!(
+                        "Adapter swap call failed: Expected amount, gas, and price elements in \
+                    the format (Uint, Uint, Tuple). Found {:?}",
+                        &return_value[..]
+                    ))),
                 }
-                _ => Err(SimulationError::FatalError("Incorrect types found for price".into())),
-            }
-        } else {
-            Err(SimulationError::FatalError("return_value is not a Token::Tuple".into()))
-        }?;
+            } else {
+                Err(SimulationError::FatalError(format!(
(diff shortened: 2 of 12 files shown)


src/protocol/vm/utils.rs
```diff
@@ -248,9 +255,9 @@ pub async fn get_code_for_contract(
     let connection_string = match connection_string {
         Some(url) => url,
         None => {
-            return Err(SimulationError::from(RpcError::InvalidRequest(
+            return Err(SimulationError::RetryDifferentInput(
```
kayibal, line 264: I think this should be Fatal no?

src/protocol/vm/utils.rs
```diff
@@ -260,22 +267,24 @@ pub async fn get_code_for_contract(
 
     // Parse the address
     let addr: H160 = address.parse().map_err(|_| {
-        SimulationError::from(RpcError::InvalidRequest(format!(
-            "Failed to parse address: {}",
-            address
-        )))
+        SimulationError::RetryDifferentInput(format!("Failed to parse address: {}", address))
```
kayibal: Unsure about this one tbh.. I would limit the try different input things to calldata - not necessary any input to any function involved in here. Tbh if a bad address is passed somewhere it is much more likely to be a bug than a temporary failure.

src/protocol/vm/utils.rs
```diff
@@ -260,22 +267,24 @@ pub async fn get_code_for_contract(
...
     })?;
 
     // Call eth_getCode to get the bytecode of the contract
     match provider.get_code(addr, None).await {
-        Ok(code) if code.is_empty() => Err(SimulationError::from(RpcError::EmptyResponse())),
+        Ok(code) if code.is_empty() => {
+            Err(SimulationError::RetryLater("Empty response from RPC".to_string()))
```
kayibal, line 283: Same here if there is no code at this address it is probably a bug.

src/protocol/vm/utils.rs
```diff
@@ -260,22 +267,24 @@ pub async fn get_code_for_contract(
...
         Ok(code) => {
             let bytecode = Bytecode::new_raw(Bytes::from(code.to_vec()));
             Ok(bytecode)
         }
         Err(e) => {
             println!("Error fetching code for address {}: {:?}", address, e);
-            Err(SimulationError::from(RpcError::InvalidResponse(e)))
+            Err(SimulationError::RetryLater(format!(
+                "Invalid response from RPC: {:?}",
+                e.to_string()
```
kayibal, line 293: Here I think it strongly depends on the returned error... e.g. if we get a 404 then it should be fatal if we get sth like 500 it may be temporary..

src/protocol/vm/adapter_contract.rs
```diff
@@ -105,15 +105,15 @@ where
                         .first()
                         .cloned()
                         .ok_or_else(|| {
-                            SimulationError::DecodingError("There wasn't a calculated price".into())
+                            SimulationError::FatalError("There wasn't a calculated price".into())
```
kayibal, line 108: This error message doesn't make any sense to me can we formulate this more clearly?

src/protocol/vm/adapter_contract.rs
```diff
@@ -105,15 +105,15 @@ where
...
                         })?;
 
                     Ok((received_amount, gas_used, price))
                 }
-                _ => Err(SimulationError::DecodingError("Incorrect types found for price".into())),
+                _ => Err(SimulationError::FatalError("Incorrect types found for price".into())),
```
kayibal, line 113: This means the adapter contract does not respect the interface right? Maybe the error message can be improved here by also returning the received types vs expected ones.

src/protocol/vm/adapter_contract.rs
```diff
@@ -199,17 +199,17 @@ where
...
                         }
                     } else {
-                        Err(SimulationError::DecodingError("Invalid fraction tuple".to_string()))
+                        Err(SimulationError::FatalError("Invalid fraction tuple".to_string()))
                     }
                 })
                 .collect()
         } else {
-            Err(SimulationError::DecodingError("Price is not a Token::Array".to_string()))
+            Err(SimulationError::FatalError("Price is not a Token::Array".to_string()))
```
kayibal, line 212: Can we print what the raw value of the price if it does not fit into the type we expected?

src/protocol/vm/state.rs
```diff
@@ -152,11 +152,11 @@ impl VMPoolState<PreCachedDB> {
                 .contains(&Capability::ScaledPrice)
             {
                 *price_result.first().ok_or_else(|| {
-                    SimulationError::DecodingError("Spot price is not a u64".to_string())
+                    SimulationError::FatalError("Spot price is not a u64".to_string())
```
kayibal, line 155: I don't think all prices will fit into a u64... I know this is outside the scope of this PR but we should probably address it...

src/protocol/vm/state_builder.rs
```diff
@@ -432,19 +428,19 @@ impl VMPoolStateBuilder {
             .map_err(|err| SimulationError::FatalError(err.to_string()))?;
 
         let address = decode(&[ParamType::Address], &sim_result.result)
-            .map_err(|_| SimulationError::DecodingError("Failed to decode ABI".into()))?
+            .map_err(|_| SimulationError::FatalError("Failed to decode ABI".into()))?
```
kayibal, line 431: It's unclear to me what this means? We weren't able to load the ABI?

src/protocol/vm/state_builder.rs
```diff
@@ -414,7 +410,7 @@ impl VMPoolStateBuilder {
 
         let parsed_address: rAddress = to_address
             .parse()
-            .map_err(|_| SimulationError::DecodingError("Invalid address format".into()))?;
+            .map_err(|_| SimulationError::FatalError("Invalid address format".into()))?;
```
kayibal, line 413: Can we add the received bad value to the error here?

src/protocol/vm/state_builder.rs
```diff
@@ -244,7 +244,7 @@ impl VMPoolStateBuilder {
             engine.state.init_account(
                 Address::parse_checksummed(to_checksum(token_address, None), None).map_err(
                     |_| {
-                        SimulationError::EncodingError(
+                        SimulationError::FatalError(
```
kayibal, line 247: Oh ok I just saw this now but I'd prefer if we would not deal with any checksummed addresses in the whole codebase. This was a mess in our cold codebase. Instead addresses should be just bytes and sometimes we represent them as a hex string - lower or uppercase should not matter.

EDIT: I realise outside of scope here again.

src/protocol/uniswap_v3/state.rs
```diff
@@ -152,11 +152,14 @@ impl UniswapV3State {
                         new_state.liquidity = state.liquidity;
                         new_state.tick = state.tick;
                         new_state.sqrt_price = state.sqrt_price;
-                        return Err(SimulationError::InsufficientData(GetAmountOutResult::new(
-                            state.amount_calculated.abs().into_raw(),
-                            gas_used,
-                            Box::new(new_state),
-                        )));
+                        return Err(SimulationError::RetryDifferentInput(
+                            "Insufficient data".into(),
```
kayibal, line 156: With tycho this is not insufficient data anymore but instead you can just call this one "ticks exceeded".

src/protocol/vm/errors.rs
```diff
@@ -23,6 +22,18 @@ pub enum FileError {
...
+        match error {
+            FileError::MalformedABI(msg) => format!("Malformed ABI error: {}", msg),
+            FileError::Structure(msg) => format!("Structure error: {}", msg),
+            FileError::FilePath(msg) => format!("File path conversion error: {}", msg),
+            FileError::Io(err) => format!("I/O error: {}", err),
+            FileError::Parse(err) => format!("Json parsing error: {}", err),
+        }
+    }
+}
+
```
louise-poole, line 36: Is this necessary? Where is this used?

src/protocol/vm/state.rs
```diff
@@ -112,8 +112,8 @@ impl VMPoolState<PreCachedDB> {
     /// Ensures the pool supports the given capability
     fn ensure_capability(&self, capability: Capability) -> Result<(), SimulationError> {
         if !self.capabilities.contains(&capability) {
-            return Err(SimulationError::NotFound(format!(
-                "capability {:?}",
+            return Err(SimulationError::FatalError(format!(
+                "capability {:?} not found",
```
louise-poole, line 116: I would word this as 'not supported' instead of 'not found'. To me 'not found' means it's not in the enum or something while 'not supported' means the pool doesn't support the capability

src/protocol/vm/state.rs
```diff
@@ -323,7 +323,7 @@ impl ProtocolSim for VMPoolState<PreCachedDB> {
         self.spot_prices
             .get(&(base.address, quote.address))
             .cloned()
-            .ok_or(SimulationError::NotFound("Spot prices".to_string()))
+            .ok_or(SimulationError::FatalError("Spot prices not found".to_string()))
```
louise-poole, line 326: 'Spot prices are empty'.
Also - is this a Fatal error? Can't the user call 'set_spot_prices' and then try again?

src/protocol/vm/state_builder.rs
```diff
@@ -195,7 +195,7 @@ impl VMPoolStateBuilder {
                 self.adapter_contract_path
                     .as_ref()
                     .ok_or_else(|| {
-                        SimulationError::NotInitialized("Adapter contract".to_string())
+                        SimulationError::FatalError("Adapter contract not initialized".to_string())
```
louise-poole, line 198: We can add more details here - it's specifically has to do with a path error no?

src/protocol/vm/state_builder.rs
```diff
@@ -386,9 +389,7 @@ impl VMPoolStateBuilder {
         let method_name = decoded
             .split(':')
             .last()
-            .ok_or_else(|| {
-                SimulationError::DecodingError("Invalid decoded string format".into())
-            })?;
+            .ok_or_else(|| SimulationError::FatalError("Invalid decoded string format".into()))?;
```
louise-poole, line 392: Can we add more detail here too? We're expecting it to be formatted as an address no?

louise-poole (commented):
Sooooo much cleaner! Love it! Left a couple nit-pick comments, for the most part this looks great!

src/protocol/vm/utils.rs
```diff
@@ -69,12 +69,19 @@ pub fn maybe_coerce_error(
...
+                "SimulationError: out-of-gas. {} Original error: {}. Pool state: {}",
+                usage_msg, data, pool_state
+            ))
         }
-        _ => err.clone(), // Otherwise return the original error
+        SimulationEngineError::TransactionError { ref data, .. } => {
+            SimulationError::FatalError(format!("TransactionError: {}", data))
+        }
+        SimulationEngineError::StorageError(message) => {
+            SimulationError::RetryLater(message.clone())
```
dianacarvalho1, line 87: why is the StorageError a RetryLater?

src/protocol/vm/adapter_contract.rs
dianacarvalho1: I would go further than @louise-poole and suggest adding which function is raising the error in all of these adapter contract errors

dianacarvalho1 (commented):
Thank you so much @tamaralipows 🙏🏼  Error handling in rust is a paaaaiiinn but this will be good for us 🤞🏼 
Only have small comments and I agree with Alan's and Louise's comments as well

src/protocol/vm/state.rs
```diff
@@ -701,9 +702,10 @@ mod tests {
         );
 
         assert!(result.is_err());
+
         match result {
             Err(e) => {
-                assert!(matches!(e, SimulationError::SellAmountTooHigh()));
+                assert!(matches!(e, SimulationError::RetryDifferentInput(_)));
```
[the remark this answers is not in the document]
tamaralipows: This is a test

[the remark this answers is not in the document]
louise-poole: I think you can do something like:
```
match result {
    Err(SimulationError::RetryDifferentInput(msg1, msg2)) => {
        assert_eq!(msg1, "expected message 1");
        assert_eq!(msg2, "expected message 2");
    }
    _ => panic!("Test failed: was expecting an Err(SimulationError::RetryDifferentInput(_, _)) value"),
}
```

src/protocol/vm/utils.rs
dianacarvalho1: oh.. oook

src/protocol/vm/utils.rs
tamaralipows: I disagree here in particular - does the user not have the option to reset their `RPC_URL` and try again?

src/protocol/vm/state.rs
tamaralipows: Are we leaving it to the user to call `set_spot_prices`? Do they know about this? That was pretty unclear to me tbh - I thought this just happened when we delta transitioned.

src/protocol/vm/utils.rs
kayibal: Do you expect them to do that while the process is running?

src/protocol/vm/state.rs
kayibal: I think this is Fatal...

src/protocol/vm/utils.rs
tamaralipows: Not sure what you mean by this, are we expecting them to change the sell amount while the process is running if the sell amount is too high? I don't really see how this would be different. This doesn't by any means indicate a bug in our own code or the protocol code - so I feel calling it Fatal would be inaccurate

src/protocol/vm/utils.rs
tamaralipows: These are the provider errors that can be raised:

AFAIU only the `JsonRpcClientError` is a 500 so I'll make a case for that one. Let me know if I'm missing something else 

```
pub enum ProviderError {
    /// An internal error in the JSON RPC Client
    #[error("{0}")]
    JsonRpcClientError(Box<dyn crate::RpcError + Send + Sync>),

    /// An error during ENS name resolution
    #[error("ens name not found: {0}")]
    EnsError(String),

    /// Invalid reverse ENS name
    #[error("reverse ens name not pointing to itself: {0}")]
    EnsNotOwned(String),

    /// Error in underlying lib `serde_json`
    #[error(transparent)]
    SerdeJson(#[from] serde_json::Error),

    /// Error in underlying lib `hex`
    #[error(transparent)]
    HexError(#[from] hex::FromHexError),

    /// Error in underlying lib `reqwest`
    #[error(transparent)]
    HTTPError(#[from] reqwest::Error),

    /// Custom error from unknown source
    #[error("custom error: {0}")]
    CustomError(String),

    /// RPC method is not supported by this provider
    #[error("unsupported RPC")]
    UnsupportedRPC,

    /// Node is not supported by this provider
    #[error("unsupported node client")]
    UnsupportedNodeClient,

    /// Signer is not available to this provider.
    #[error("Attempted to sign a transaction with no available signer. Hint: did you mean to use a SignerMiddleware?")]
    SignerUnavailable,
}
```

src/protocol/vm/utils.rs
louise-poole: I think this should be Fatal. An RPC url is a set-up issue, not a runtime issue... it shouldn't ever change while the program is running and needs external input by a human in order to fix it if it is wrong.
As for sell amount errors - yes, that's a runtime error and can easily be handled automatically by a program that recognises the error and reduces the sell amount to try again. This doesn't need human input to correct.

louise-poole (approved):
LGTM!

PR #70, opened 2024-11-27
feat: make most VMPoolState attrs private

tamaralipows:
- The only one that is actually ever accessed outside the VMPoolState is `tokens` (tycho.rs, line 364), so I'm leaving it as is for now until any possible tycho.rs refactors. Any other attributes were only accessed inside tests, so I made getters that only compile at test time.
- Note: The USV3 pool state has attrs private, while the USV2 pool state has its attrs public. Not sure the reasoning behind this.
- This PR didn't take any time at all so don't be afraid to fight me if I missed some reasoning and you think this is bad. It can be closed.


dianacarvalho1 (approved):
>  so I made getters that only compile at test time.

Oh my, you are a rust pro already 💅🏼 
Thank you so much for this 🙏🏼 

I'm not sure about v2 and v3 decisions though..
What we could argue for is: the user might want to do wtv weird things with these pool states (outside of our simulations) and they might need to access some of the fields?

src/protocol/vm/state.rs
```diff
@@ -43,36 +43,36 @@ where
...
-    pub block: BlockHeader,
+    block: BlockHeader,
     /// The pools token balances
-    pub balances: HashMap<H160, U256>,
+    balances: HashMap<H160, U256>,
     /// The contract address for where protocol balances are stored (i.e. a vault contract).
     /// If given, balances will be overwritten here instead of on the pool contract during
     /// simulations
-    pub balance_owner: Option<H160>,
+    balance_owner: Option<H160>,
```
louise-poole, line 56: Question: why not make this pub(super)? Then it can be accessed within this `vm` module as needed, but is considered private outside of the module.

tamaralipows: hehe didn't know it existed. but is it necessary since it's really only ever accessed for tests?

PR #75, opened 2024-11-28
fix: Don't use unwrap_or_default if balance not found

tamaralipows:
This is bug prone. If the user forgets to initialize the balance, we should raise an error instead.

### DIFF
--- a/src/evm/protocol/vm/state_builder.rs
+++ b/src/evm/protocol/vm/state_builder.rs
@@ -204,7 +204,12 @@ impl EVMPoolStateBuilder {
             self.id,
             self.tokens,
             self.block,
-            self.balances.unwrap_or_default(),
+            self.balances.ok_or_else(|| {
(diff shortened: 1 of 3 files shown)


src/evm/protocol/vm/state_builder.rs
```diff
@@ -204,7 +204,12 @@ impl EVMPoolStateBuilder {
             self.id,
             self.tokens,
             self.block,
-            self.balances.unwrap_or_default(),
+            self.balances.ok_or_else(|| {
```
zizou0x, line 207: Ok I wasn't sure about this in my PR. Let's move it to the 'new' function and make it not Option<> instead? 

That way it would be clearer from the interfaces. In general it's better to error at compile time than at runtime when possible.

zizou0x (approved):
Thanks! Looks good, just one comment

PR #76, opened 2024-11-28
feat: Discontinue support for event transitioning

tamaralipows:
This is legacy code, and we are aiming to keep the repo as simple as possible.


dianacarvalho1 (approved):
🧹 ✨ Thaaankk yooouu 🙏🏼 I love me a good cleaann up.
And with all the refactors we had already done, this was pretty smooth, very good very good!

Just one question: can't we get rid of the ProtocolEvent as well?

louise-poole (approved):
I agree with @zizou0x , the code itself looks good though!

zizou0x (approved):

PR #77, opened 2024-12-02
feat: take generic constraints for EVMPoolState instead of PreCachedDB

tamaralipows:
I'm not so familiar with thread safety in rust but I've tried my best. Please be patient but give harsh critiques where necessary 🙏

- This is in preparation for adding a test using `SimulationDB` instead of `PreCachedDB`
- The Send and Sync are required by the trait bounds of ProtocolSim. We otherwise get the error "`D` cannot be shared between threads safely"
- For this, we needed to change the account_storage in the SimulationDB to use `Arc<RwLock<AccountStorage>>` instead of `RefCell`. Otherwise we would get unsatisfied trait bounds due to the generic trait requiring Send and Sync.
- About the `unwrap()` after all the `.read()`s and `.write()`s: I understand it's bad practice to use `unwrap()` in production code, but I see this used several times in `tycho_db.rs` so I am assuming this is an exception.


src/evm/engine_db/simulation_db.rs
```diff
@@ -94,7 +97,7 @@ pub struct SimulationDB<M: Middleware> {
     /// Client to connect to the RPC
     client: Arc<M>,
     /// Cached data
-    account_storage: RefCell<AccountStorage>,
+    account_storage: Arc<RwLock<AccountStorage>>,
```
dianacarvalho1, line 100: by doing this I think we are trading off some performance for thread safety and generality 😕 
Do we think the performance decrease will be significant? @zizou0x maybe you have some idea here? (because I have none 😅 )

dianacarvalho1 (commented):
Thank you sooo much for this @tamaralipows 🙏🏼 
I also don't know if this is a good trade off or not though 😕 waiting for some more experienced rustaceans 🦀 to have a look

src/evm/engine_db/simulation_db.rs
kayibal: This is the correct way to do it!

zizou0x: Agreed with @kayibal, you're right that it adds some overhead but in that particular case the bottleneck will be network by far, and anyone who use this shouldn't expect high performances, so we can afford this IMO.

zizou0x (approved):
Looks good to me, thanks!

You're right that in many cases we want to avoid `unwrap`. In that case it's not a normal `Result` but a [LockResult](https://doc.rust-lang.org/std/sync/type.LockResult.html) which error only if the lock is poisoned (meaning a thread panicked with the lock). So it's indeed a particular case and I think unwrapping here is good.

PR #78, opened 2024-12-03
fix: Don't copy block_lasting_overwrites in get_overwrites

dianacarvalho1:
This was breaking the defibot tests

The whole block_lasting_overwrites logic relies on being able to change the object in place 🫠


tamaralipows (approved):

tamaralipows:
on hold until python is important or we have time

PR #82, opened 2024-12-04
feat: Take generic DB trait in state builder 

tamaralipows:
This PR was an attempt to add a test for `SimulationDB`. It did not go well.
I've left the test there so you can see what I've attempted, but I'd like to remove it before merging.

However, thanks to this task, I've cleaned up a lot of issues that we missed in the state builder, that were necessary setup in order to be able to run this with the `SimulationDB` _eventually_. These issues were also making our code inconsistent (i.e. we were using generic traits some places, and hardcoding `PreCachedDB` in others).


src/evm/protocol/vm/state.rs
```diff
@@ -599,11 +604,107 @@ mod tests {
...
+        let data: Value = serde_json::from_str(data_str).expect("Failed to parse JSON");
+
+        let accounts: Vec<AccountUpdate> = serde_json::from_value(data["accounts"].clone())
+            .expect("Expected accounts to match AccountUpdate structure");
+
+        let rpc_url = std::env::var("ETH_RPC_URL").unwrap();
+        let client = Provider::<Http>::try_from(rpc_url).unwrap();
+        let runtime = Arc::new(tokio::runtime::Runtime::new().unwrap());
+        let client = Arc::new(client);
+        let db = SimulationDB::new(client, None, None);
```
tamaralipows, line 629: We should be passing the runtime here. Unfortunately we can't do that without moving it, and we need this very runtime below in order to build the pool state.

When we don't pass the runtime here, we the problems:
```
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

src/evm/protocol/vm/state.rs
```diff
@@ -599,11 +604,107 @@ mod tests {
...
+            .adapter_contract_path(PathBuf::from(
+                "src/evm/protocol/vm/assets/BalancerSwapAdapter.evm.runtime".to_string(),
+            ))
+            .stateless_contracts(stateless_contracts)
+            .engine(engine)
+            .build(db);
+
+        let pool_state = runtime
+            .block_on(build_future)
+            .expect("Failed to build pool state");
```
tamaralipows, line 688: This was the original issue - the `build()` method is async.

src/evm/protocol/vm/state.rs
```diff
@@ -507,15 +507,20 @@ mod tests {
         collections::{HashMap, HashSet},
         path::PathBuf,
         str::FromStr,
+        sync::Arc,
     };
 
-    use ethers::{prelude::H256, types::Address as EthAddress};
+    use ethers::{
+        prelude::{Http, Provider, H256},
+        types::Address as EthAddress,
```
dianacarvalho1, line 515: 👀 👀 👀 👀 👀 👀 👀  eettheeeerrsssss?? I banneeedd yooouuu! you shall not return to this repo!!

dianacarvalho1: kidding, I will remove it again after this is merged 😂

src/evm/protocol/vm/state_builder.rs
```diff
@@ -87,12 +92,17 @@ pub struct EVMPoolStateBuilder {
     token_storage_slots: Option<HashMap<H160, (ERC20Slots, ContractCompiler)>>,
     manual_updates: Option<bool>,
     trace: Option<bool>,
-    engine: Option<SimulationEngine<PreCachedDB>>,
-    adapter_contract: Option<TychoSimulationContract<PreCachedDB>>,
+    engine: Option<SimulationEngine<D>>,
+    adapter_contract: Option<TychoSimulationContract<D>>,
```
dianacarvalho1, line 96: oh my god, this looks perfect now tbh!

dianacarvalho1 (approved):
I love it ❤️ 
I know we didn't achieve the ultimate goal of the task but I reaalllyy like this stepping stone! It really is cleaning up the code and making things nicer 🙏🏼 
We can try to come back to this later (if we think it's important enough)

PR #83, opened 2024-12-05
feat: Remove ethers from interfaces and uniswap v2 and v3 🧹 

dianacarvalho1:
tldr: Using BigUInt for interfaces and alloy for EVM related code

Changes:
- Add num-bigint package
- Change GetAmountOutResult to use BigUint
- In get_amount_out change amount_in to be BigUint
- Change all_tokens in try_from_with_block to be Bytes and not H160
- ERC20Token uses BigUint for gas
- Uniswap v2, v3  and VM: changed get_amount_out signature to use BigUint 
- Uniswap v2 and v3 updated to use alloy instead of ethers
- Update example to new interface changes

Clean up:
- Delete EVMLogMeta and related code - it isn't used anywhere.
- Remove LogIndex from UniswapV2 and UniswapV3
- Delete Pair and Swap structs - this isn't used anywhere. It is used in the token quoter but we can construct it there later
- Update doc examples

TODO:
- ERC20 still uses ethers and it is used in a lot of interfaces. We need to abstract this away (separate PR coming next)
- Change VM and engine code to use alloy (I gave it a small shot already and 🥴 I think it's low prio now though and will come back to it later in the mission/cooldown)


Cargo.toml
```diff
@@ -43,18 +43,19 @@ tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", p
...
-foundry-config = { git = "https://github.com/foundry-rs/foundry", rev = "57bb12e",  optional = true }
-foundry-evm = { git = "https://github.com/foundry-rs/foundry", rev = "57bb12e",  optional = true }
+foundry-config = { git = "https://github.com/foundry-rs/foundry", rev = "57bb12e", optional = true }
+foundry-evm = { git = "https://github.com/foundry-rs/foundry", rev = "57bb12e", optional = true }
 alloy-primitives = { version = "0.8.9", features = [
     "getrandom",
     "rand",
     "map-foldhash",
-], optional = true }
+] }
```
kayibal, line 55: This will eventually become optional, right?

src/u256_num.rs
kayibal, line 1: should we move this under `evm/protocol`?

src/safe_math.rs
kayibal, line 1: Should we move this under `evm/protocol`

src/evm/protocol/uniswap_v3/state.rs
```diff
@@ -248,12 +249,16 @@ impl ProtocolSim for UniswapV3State {
...
-        amount_in: U256,
+        amount_in: BigUint,
         token_a: &ERC20Token,
         token_b: &ERC20Token,
     ) -> Result<GetAmountOutResult, SimulationError> {
         let zero_for_one = token_a < token_b;
-        let amount_specified = I256::checked_from_sign_and_abs(Sign::Positive, amount_in).unwrap();
+        let amount_specified = I256::checked_from_sign_and_abs(
+            Sign::Positive,
+            U256::from_be_slice(&amount_in.to_bytes_be()),
```
kayibal, line 259: Unfortunately this allocates a Vec<u8> and immediately references it... this is not great but I couldn't find a better way to do it...

kayibal (approved):
Gave this a quick review and so far looks good. The change toi alloy-rs makes this a bit more difficult to review.

Cargo.toml
dianacarvalho1: why?? it will be necessary for the main interfaces

src/evm/protocol/uniswap_v2/reserve_price.rs
```diff
@@ -1,18 +1,17 @@
-use ethers::types::U256;
-
 use crate::u256_num::u256_to_f64;
+use alloy_primitives::U256;
```
tvinagre, line 2: Sorry if it's a stupid question, but in some places you've introduced BigUInt - and here, Alloy U256 - what's the reason? Capture overflows the same way it'd happen on-chain?

src/evm/protocol/uniswap_v3/solidity_math.rs
```diff
@@ -1,31 +1,39 @@
...
+    // Access the limbs of the U512 value
+    let limbs = value.as_limbs();
+
+    // Check if the upper 256 bits are non-zero
+    if limbs[4] != 0 || limbs[5] != 0 || limbs[6] != 0 || limbs[7] != 0 {
+        return Err(SimulationError::FatalError("Overflow: Value exceeds 256 bits".to_string()));
+    }
+
+    // Extract the lower 256 bits
+    Ok(U256::from_limbs([limbs[0], limbs[1], limbs[2], limbs[3]]))
```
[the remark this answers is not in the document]
dianacarvalho1: because ethers did this under the hood and alloy doesn't. we have to explicitly define this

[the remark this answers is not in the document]
tamaralipows: Did you just copy this from the ethers lib then?

src/evm/protocol/uniswap_v2/reserve_price.rs
tvinagre: If I understand correctly, Alloy for EVM-related code and BigUint for Tycho types

tvinagre (commented):
Looks good! A few questions, but nothing major. Tiny comment, I feel like this PR does too much by also removing unused code.

src/evm/protocol/uniswap_v2/reserve_price.rs
dianacarvalho1: yes exactly! BigUInt for interfaces and alloy for EVM related code! (Adding this in the PR description as well)

tamaralipows (approved):
Thank you! All makes sense :)

PR #84, opened 2024-12-05
feat: Change ERC20Token address to be alloy's Address instead of H160

dianacarvalho1:
Also change the VMPoolState attributes to be Address instead of H160

TODO: (in the upcoming episodes):
- Replace ethers U256 with alloy U256 in the vm pool state
- Abstract away ERC20Token logic (this one is evm specific)
- Use alloy to encode and decode instead of ethers (in simulation contract)

### DIFF
--- a/src/evm/protocol/vm/adapter_contract.rs
+++ b/src/evm/protocol/vm/adapter_contract.rs
@@ -1,10 +1,11 @@
+use alloy_primitives::Address;
 use std::collections::{HashMap, HashSet};
 
 use ethers::{
-    abi::{Address, Token},
-    types::U256,
+    abi::Token,
+    types::{H160, U256},
 };
-use revm::{primitives::Address as rAddress, DatabaseRef};
+use revm::DatabaseRef;
 
 use crate::{
     evm::{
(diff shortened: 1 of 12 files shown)


src/evm/protocol/vm/erc20_token.rs
```diff
@@ -138,39 +139,33 @@ lazy_static! {
...
 {
-    let token_contract = TychoSimulationContract::new(
-        rAddress::from_slice(token_addr.as_bytes()),
-        engine.clone(),
-        ERC20_ABI.clone(),
-    )
-    .unwrap();
-
-    let external_account = H160::from_slice(&*EXTERNAL_ACCOUNT.0);
+    let token_contract =
```
tamaralipows, line 150: here

tamaralipows (approved):
I love it! *chef's kiss*

dianacarvalho1:
@tamaralipows needs this in her current task so we decided to merge right away. Feel free to review it and I will address the comments in the next PR (which will be in the same scope anyway)

src/evm/protocol/vm/adapter_contract.rs
```diff
@@ -51,12 +52,12 @@ where
...
         amounts: Vec<U256>,
         block: u64,
-        overwrites: Option<HashMap<rAddress, Overwrites>>,
+        overwrites: Option<HashMap<Address, Overwrites>>,
     ) -> Result<Vec<f64>, SimulationError> {
         let args = vec![
             Token::FixedBytes(pair_id),
-            Token::Address(sell_token),
-            Token::Address(buy_token),
+            Token::Address(H160(sell_token.0 .0)),
```
tvinagre, line 59: Can't we create Address instead of going through H160? This keeps the `ethers` dependency that we're trying to get rid of

dianacarvalho1: yes I know.. I'm trying to do these in the smallest parts I can so it's not crazy overwhelming. 😓 
This part will be addressed in the last point of the TODOs in the description
> Use alloy to encode and decode instead of ethers (in simulation contract)

PR #85, opened 2024-12-05
feat: Generalize ERC20Token class

tamaralipows:
- ERC20Token was used in instances that were not EVM-specific, which is bad organization.
- We decided that creating a token generic trait was overkill - especially since the main issue was just the Bytes. Instead, we/I decided that a good and simple solution was to:

1. Rename ERC20Token -> Token
2. Make the address Bytes instead of Address
3. Create a helper method to transform the Bytes to Address when needed


src/protocol/models.rs
```diff
@@ -63,7 +63,7 @@ pub trait TryFromWithBlock<T> {
     async fn try_from_with_block(
         value: T,
         block: Header,
-        all_tokens: HashMap<Address, ERC20Token>,
+        all_tokens: HashMap<Address, Token>,
```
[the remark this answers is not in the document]
tamaralipows: Good catch!!

src/models.rs
```diff
@@ -1,25 +1,25 @@
...
 //!
 //! This module contains basic models that are shared across many
-//! components of the crate, including ERC20Token, Swap and SwapSequence.
+//! components of the crate, including Token, Swap and SwapSequence.
 //!
-//! ERC20Tokens provide instructions on how to handle prices and amounts,
+//! Tokens provide instructions on how to handle prices and amounts,
 //! while Swap and SwapSequence are usually used as results types.
-use alloy_primitives::{Address, U256};
+use alloy_primitives::{bytes::Bytes, U256};
```
zizou0x, line 8: Shouldn't we use our own `tycho_core::Bytes`?

src/evm/protocol/erc20.rs
```diff
@@ -0,0 +1,38 @@
...
+pub fn bytes_to_erc20_address(address: &Bytes) -> Result<Address, SimulationError> {
+    if address.len() == 20 {
+        Ok(Address::from_slice(address))
+    } else {
+        Err(SimulationError::InvalidInput(
+            format!("Invalid ERC20 token address: {:?}", address),
+            None,
+        ))
+    }
+}
```
zizou0x, line 17: I think we should consider making a trait for this in `tycho-ethereum` (like we did for ethers type [here](https://github.com/propeller-heads/tycho-indexer/blob/82f9974a864a5d4ae7f216e64b74c5f73c020229/tycho-ethereum/src/lib.rs#L51-L127)) so we can reuse it, for example in the token-quoter repo

src/evm/protocol/erc20.rs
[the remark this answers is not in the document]
tamaralipows: nope because its also used for usv2 and 3

src/evm/protocol/erc20.rs
tamaralipows: Nice idea - not urgent for the release though I guess?

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 looks good to me!

examples/explorer/main.rs
```diff
@@ -39,31 +39,21 @@ async fn main() {
...
-    // let (tick_tx, mut tick_rx) = mpsc::channel::<BlockState>(12);
-
-    // let consumer_handle = task::spawn(async move {
-    //     while let Some(message) = tick_rx.recv().await {
-    //         println!("got message: {:?}", message);
-    //     }
-    //     Ok(())
-    // });
-    // let tasks = [tycho_message_processor, consumer_handle];
+    let tasks = [tycho_message_processor, consumer_handle];
```
tamaralipows, line 56: Oops. Reverting this file. Ignore this.

zizou0x (approved):

PR #87, opened 2024-12-07
test: Sequential get_amount_out calls

tamaralipows:
This is equivalent to `test_sequential_get_amount_outs` on the python side - the most important assertion being that the spot price does not remain the same after the first swap.


dianacarvalho1 (approved):
awessoomeeee! Thank you 🙏🏼 
and that our overwrites make sense and no weird things are happening in rust like it is in python!

zizou0x (approved):

louise-poole (approved):

PR #88, opened 2024-12-09
feat: Replace ethers Middleware with Provider from alloy

dianacarvalho1:
ethers is **gone** from everywhere

(review commit by commit for more info on decisions)


src/evm/engine_db/simulation_db.rs
```diff
@@ -242,22 +230,26 @@ where
...
+                    .get_storage_at(address, index)
+                    .number(block.number)
+                    .await
+                    .unwrap()
+            } else {
+                self.client
+                    .get_storage_at(address, index)
+                    .await
+                    .unwrap()
+            }
```
tamaralipows, line 252: I am a bit confused about the implications of this change, why are we still using `block_on`? Is that necessary now with an async provider? Or will we just address this in a future PR?

```
    fn block_on<F: core::future::Future>(&self, f: F) -> F::Output {
        // If we get here and have to block the current thread, we really
        // messed up indexing / filling the storage. In that case this will save us
        // at the price of a very high time penalty.
        match &self.runtime {
            Some(runtime) => runtime.block_on(f),
            None => futures::executor::block_on(f),
        }
    }

File: src/evm/engine_db/simulation_db.rs [hunk 2]
```diff
@@ -322,11 +314,11 @@ where
...
 }
 
-impl<M: Middleware + Debug> DatabaseRef for SimulationDB<M>
+impl<P: Provider + Debug> DatabaseRef for SimulationDB<P>
 where
-    M::Error: std::fmt::Debug,
+    P: Provider + Send + Sync + 'static,
 {
-    type Error = M::Error;
+    type Error = Box<dyn std::error::Error + Send + Sync>;
```
tamaralipows, line 322: After all we needed the lifetimes I guess? Thanks for your bravery

tamaralipows (commented):
Thanks a lot for this - I have one main question regarding the `block_on`, but I'm very happy you managed to make this work without ethers! [️‍

src/evm/engine_db/simulation_db.rs
dianacarvalho1: yeah I think we confused ourselves. ethers Middleware was async as well. (See [here](https://docs.rs/ethers-providers/2.0.14/src/ethers_providers/middleware.rs.html#147)). it's the DatabaseRef that isn't async. so we need to wrap the asyncness in the block_on to be sync.
when the async databaseref is release we can change our code to use it and remove this block_on!

src/evm/engine_db/simulation_db.rs
```diff
@@ -105,12 +98,9 @@ pub struct SimulationDB<M: Middleware + Debug> {
     pub runtime: Option<Arc<tokio::runtime::Runtime>>,
 }
 
-impl<M: Middleware + Debug> SimulationDB<M>
-where
-    M::Error: std::fmt::Debug,
-{
+impl<P: Provider + Debug + 'static> SimulationDB<P> {
```
zizou0x, line 101: Why do we need it to be `'static`?

src/evm/engine_db/simulation_db.rs
```diff
@@ -322,11 +314,11 @@ where
...
 }
 
-impl<M: Middleware + Debug> DatabaseRef for SimulationDB<M>
+impl<P: Provider + Debug> DatabaseRef for SimulationDB<P>
 where
-    M::Error: std::fmt::Debug,
+    P: Provider + Send + Sync + 'static,
 {
-    type Error = M::Error;
+    type Error = Box<dyn std::error::Error + Send + Sync>;
```
zizou0x, line 321: Can't we make this not `dyn` but a known type instead?

src/evm/engine_db/simulation_db.rs
```diff
@@ -322,11 +314,11 @@ where
     }
 }
 
-impl<M: Middleware + Debug> DatabaseRef for SimulationDB<M>
+impl<P: Provider + Debug> DatabaseRef for SimulationDB<P>
 where
-    M::Error: std::fmt::Debug,
+    P: Provider + Send + Sync + 'static,
```
zizou0x, line 319: Better to keep everything together, also `Provider` is specified twice

```suggestion
impl<P> DatabaseRef for SimulationDB<P>
where
    P: Provider + Debug + Send + Sync + 'static,
```

And same question as above: what forces us to make it `'static`?

zizou0x (commented):
Looks 🔥 thanks! 

Just have some questions to better understand the generics and lifetimes and make sure this is the best we can do

src/evm/engine_db/simulation_db.rs
dianacarvalho1: huumm the type depends on the provider I think. that's why I didn't want to define it here

src/evm/engine_db/simulation_db.rs
dianacarvalho1: if I don't, I get

```
error[E0310]: the parameter type `P` may not live long enough
   --> src/evm/engine_db/simulation_db.rs:270:55
    |
270 | impl<P: Provider + Debug> EngineDatabaseInterface for SimulationDB<P>
    |                                                       ^^^^^^^^^^^^^^^
    |                                                       |
    |                                                       the parameter type `P` must be valid for the static lifetime...
    |                                                       ...so that the type `P` will meet its required lifetime bounds...
    |
note: ...that is required by this bound
   --> src/evm/engine_db/engine_db_interface.rs:6:36
    |
6   | pub trait EngineDatabaseInterface: DatabaseRef + Send + Sync {
    |                                    ^^^^^^^^^^^
help: consider adding an explicit lifetime bound
    |
270 | impl<P: Provider + Debug + 'static> EngineDatabaseInterface for SimulationDB<P>
    |                          +++++++++
```

src/evm/engine_db/simulation_db.rs
tamaralipows: ahhhhh ok nice

tamaralipows (approved):

src/evm/engine_db/simulation_db.rs
```diff
@@ -199,38 +189,36 @@ where
...
-                    .unwrap_or_else(|e| panic!("ethers get balance error: {e:?}"))
-                    .0,
-            ),
-            nonce
-                .unwrap_or_else(|e| panic!("ethers get nonce error: {e:?}"))
-                .as_u64(),
-            code.hash_slow(),
-            code,
-        ))
+        let code = to_analysed(Bytecode::new_raw(revm::primitives::Bytes::copy_from_slice(&code?)));
```
tvinagre, line 219: uf - so much cleaner

src/evm/engine_db/simulation_db.rs
```diff
@@ -683,38 +619,29 @@ mod tests {
...
-        mock_sim_db.init_account(
-            address1,
-            AccountInfo::default(),
-            Some(original_storage.clone()),
-            false,
-        );
+        db.init_account(address1, AccountInfo::default(), Some(original_storage.clone()), false);
         let address2 = Address::from_str("0000000000000000000000000000000000000002").unwrap();
-        mock_sim_db.init_account(address2, AccountInfo::default(), Some(original_storage), false);
+        db.init_account(address2, AccountInfo::default(), Some(original_storage), false);
```
tvinagre, line 639: For improved readability:
```
    // Setup addresses
    let address1 = Address::from_str("0000000000000000000000000000000000000001").unwrap();
    let address2 = Address::from_str("0000000000000000000000000000000000000002").unwrap();
    let address3 = Address::from_str("0000000000000000000000000000000000000003").unwrap();

    // Initialize accounts
    db.init_account(address1, AccountInfo::default(), Some(original_storage.clone()), false);
    db.init_account(address2, AccountInfo::default(), Some(original_storage), false);
```

src/evm/protocol/vm/erc20_token.rs
```diff
@@ -305,13 +309,28 @@ mod tests {
         assert_eq!(overwrites[&factory.token_address][&factory.total_supply_slot], supply);
     }
 
+    fn new_state() -> SimulationDB<RootProvider<BoxTransport>> {
+        let eth_rpc_url = env::var("ETH_RPC_URL").unwrap_or_else(|_| {
+            dotenv().expect("Missing .env file");
+            env::var("ETH_RPC_URL").expect("Missing ETH_RPC_URL in .env file")
```
tvinagre, line 315: This is kinda weird and doesn't it check for the envvar twice?
```
    dotenv().ok();  // first we should load dotenv and return ok() because it's fine if .env doesn't exist in tests IMO
    
    // Then check for env var once
    let eth_rpc_url = env::var("ETH_RPC_URL")
        .expect("Missing ETH_RPC_URL in environment");
```

src/evm/simulation.rs
```diff
@@ -601,30 +605,43 @@ mod tests {
...
     }
-
-    #[test]
-    fn test_integration_revm_v2_swap() -> Result<(), Box<dyn Error>> {
-        let client = Provider::<Http>::try_from(std::env::var("ETH_RPC_URL").unwrap()).unwrap();
-        let client = Arc::new(client);
+    fn new_state() -> SimulationDB<RootProvider<BoxTransport>> {
+        let eth_rpc_url = env::var("ETH_RPC_URL").unwrap_or_else(|_| {
+            dotenv().expect("Missing .env file");
+            env::var("ETH_RPC_URL").expect("Missing ETH_RPC_URL in .env file")
```
tvinagre, line 626: Same comment as previous test

src/evm/protocol/vm/erc20_token.rs
```diff
@@ -305,13 +309,28 @@ mod tests {
...
+            .is_err()
+            .then(|| tokio::runtime::Runtime::new().unwrap())
+            .unwrap();
+        let client = runtime.block_on(async {
+            ProviderBuilder::new()
+                .on_builtin(&eth_rpc_url)
+                .await
+                .unwrap()
+        });
+        let client = Arc::new(client);
```
tvinagre, line 327:
```suggestion
SimulationDB::new(Arc::new(client), Some(Arc::new(runtime)), None)
```

tvinagre (approved):
Looks good! Just some small suggestions

PR #89, opened 2024-12-09
docs: In-code docs cleanup

tamaralipows:
We have decided to focus on documenting the public interfaces for now - the private ones remain mostly undocumented.

**Main points:**
- Document methods `ensure_capability` and `set_spot_prices`
- Fix all mentions `TradeSimulationError` - this no longer exists.
- Remove redundant docs in implementations of `ProtocolSim` (keep the main trait docs as the ground truth)
- Remove `block_lasting_ovewrites`
- Attempt `Capability` docs (see TODOs)

**Fully double-checked:**
 - `protocol/state.rs`
 - `protocol/*/state.rs`
 - `protocol/*/tycho_decoders.rs`


src/evm/protocol/vm/models.rs
```diff
@@ -3,6 +3,45 @@ use strum_macros::Display;
...
 use crate::protocol::errors::SimulationError;
 
+/// Represents a distinct functionality or feature that an `EVMPoolState` can support.
+///
+/// Each `Capability` variant corresponds to a specific functionality that influences how the
+/// simulation interacts with the `EVMPoolState`.
+///
+/// # Variants
+///
+/// TODO I don't know what SellSide or BuySide mean (GPT-generated)
```
dianacarvalho1, line 13: I think capabilities are documented [here](https://github.com/propeller-heads/tycho-protocol-sdk/blob/main/evm/src/interfaces/ISwapAdapterTypes.sol#L15)

src/evm/protocol/vm/models.rs
```diff
@@ -3,6 +3,45 @@ use strum_macros::Display;
...
+/// use crate::protocol::errors::SimulationError;
+///
+/// let capability = Capability::SellSide;
+///
+/// match capability {
+///     Ok(Capability::SellSide) => println!("Supports sell-side operations."),
+///     Ok(cap) => println!("Other capability: {:?}", cap),
+///     Err(err) => println!("Error: {}", err),
+/// }
+/// ```
```
dianacarvalho1, line 44: I think this is redundant no?

src/evm/protocol/vm/state.rs
```diff
@@ -126,6 +135,34 @@ where
         Ok(())
     }
 
+    /// Sets the spot prices for a pool for all possible pairs of the given tokens.
```
dianacarvalho1, line 138: maybe here we can also add some advice on usage?
like that if the use is interested in fetching really fast prices it would be nice to set the spot prices on the pool every time the pool actually changes and not only when the user needs the prices

src/evm/protocol/vm/state.rs
```diff
@@ -172,9 +209,19 @@ where
         Ok(())
     }
 
-    /// Retrieves the sell amount limit for a given pair of tokens, where the first token is treated
-    /// as the sell token and the second as the buy token. The order of tokens in the input vector
-    /// is significant and determines the direction of the price query.
+    /// Retrieves the sell amount limit for a given pair of tokens and the given overwrites.
```
dianacarvalho1, line 212: in the end.. what is a sell mount limit?? a limit sell amount that the pool can trade? maybe we can make that more obvious here

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 I have small suggestions

src/evm/protocol/vm/state.rs
```diff
@@ -653,6 +653,7 @@ mod tests {
         db.update(accounts, Some(block));
 
         let tokens = vec![dai().address, bal().address];
+        println!("{:?}", tokens);
```
dianacarvalho1, line 656: 👀

dianacarvalho1 (approved):
Aweesoomeeee ✨ tiny tiny comment! Thank yoouuu

PR #90, opened 2024-12-10
chore: update default TYCHO_URL to point to beta environment

zizou0x:
After the recent Tycho prod update, we can point to beta endpoint again


dianacarvalho1 (approved):
marvellous ✨

PR #91, opened 2024-12-10
feat(example): add Curve to `explorer` example

zizou0x:
The purpose of this PR is to add Curve support in the `explorer` example. 

It also contains a few small fixes that are needed to support Curve.

### DIFF
--- a/src/evm/protocol/vm/utils.rs
+++ b/src/evm/protocol/vm/utils.rs
@@ -355,19 +355,49 @@ pub fn hexstring_to_vec(hexstring: &str) -> Result<Vec<u8>, SimulationError> {
     Ok(bytes)
 }
 
+/// Converts a hexadecimal string into a fixed-size 32-byte array.
+///
+/// This function takes a string slice (e.g., a pool ID) that may or may not have
+/// a `0x` prefix. It decodes the hex string into bytes, ensuring it does not exceed
+/// 32 bytes in length. If the string is valid and fits within 32 bytes, the bytes
+/// are copied into a `[u8; 32]` array, with right zero-padding for unused bytes.
+///
+/// # Arguments
+///
+/// * `pool_id` - A string slice representing a hexadecimal pool ID. It can optionally start with
+///   the `0x` prefix.
+///
+/// # Returns
+///
+/// * `Ok([u8; 32])` - On success, returns a 32-byte array with the decoded bytes. If the input is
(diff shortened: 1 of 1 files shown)


src/evm/protocol/vm/state.rs
```diff
@@ -219,13 +219,13 @@ where
         tokens: Vec<Address>,
         max_amount: U256,
     ) -> Result<HashMap<Address, Overwrites>, SimulationError> {
-        let sell_token = &tokens[0].clone();
+        let sell_token = &tokens[0].clone(); //TODO: need to make it clearer from the interface
```
zizou0x, line 222: I don't think this PR is the correct place to change this so I added a TODO.

src/evm/protocol/vm/state_builder.rs
```diff
@@ -253,8 +253,8 @@ where
 
         if let Some(stateless_contracts) = &self.stateless_contracts {
             for (address, bytecode) in stateless_contracts.iter() {
+                let mut addr_str = address.clone();
```
dianacarvalho1, line 256: ah shit, good catch! 🙏🏼

src/evm/protocol/vm/utils.rs
```diff
@@ -350,22 +355,161 @@ pub fn hexstring_to_vec(hexstring: &str) -> Result<Vec<u8>, SimulationError> {
...
+/// ```
+/// use json_deserialize_address_list;
+///
+/// let json_input = br#"["0x1234", "0xc0ffee"]"#;
+/// match json_deserialize_address_list(json_input) {
+///     Ok(result) => println!("Decoded: {:?}", result),
+///     Err(e) => eprintln!("Error: {}", e),
+/// }
+/// ```
+pub fn json_deserialize_address_list(input: &[u8]) -> Result<Vec<Vec<u8>>, SimulationError> {
```
dianacarvalho1, line 434: where is this being used??

dianacarvalho1 (approved):
Thank you so much @zizou0x 🙏🏼 only have 1 small question

src/evm/protocol/vm/utils.rs
zizou0x: It's the deserialize function related to [this serialize function](https://github.com/search?q=repo:propeller-heads/tycho-protocol-sdk%20json_serialize_address_list&type=code) that we have in the SDK. 

Currently it's only used by Balancer for rate providers.

PR #94, opened 2024-12-12
fix: Fix adapter contract hardcoded address

dianacarvalho1:
Don't hardcode ADAPTER_ADDRESS, make it depend on the protocol


tamaralipows (approved):
Thank you!

PR #96, opened 2024-12-13
docs: Update Readmes

dianacarvalho1:
Small bonus refactoring: Rename bytes_to_erc20_address -> bytes_to_address


Readme.md
```diff
@@ -2,26 +2,20 @@
...
 > documentation.
 
 ## Currently supported protocols:
 
 - Uniswap V2 and Forks
 - Uniswap V3
-- VM enabled protocols
-
-See also `evm` module which can simulate any transaction.
+- VM enabled protocols: Balancer v2 and Curve
```
tamaralipows, line 18: sry being picky 

```suggestion
- VM enabled protocols: Balancer V2 and Curve
```

Readme.md
```diff
@@ -30,21 +24,21 @@ To add a new protocol, you will need to complete the following high-level steps:
...
 
 Each native protocol should have its own module under `tycho-simulation/src/protocol`.
 
 #### VM protocol:
 
 1. Add the associated adapter runtime file to `tycho-simulations/src/protocol/assets`. Make sure to name the file
-according to the protocol name used by tycho in the following format: `<Protocol><Version>Adapter.evm.runtime`. 
-For example: `vm:balancer_v2` will be `BalancerV2Adapter.evm.runtime`.
+   according to the protocol name used by tycho in the following format: `<Protocol><Version>Adapter.evm.runtime`.
+   For example: `vm:balancer_v2` will be `BalancerV2Adapter.evm.runtime`.
```
tamaralipows, line 36: Should we at some point link to the gitbook docs?
This piece of information is pretty much useless without knowing how to generate this runtime file

src/evm/Readme.md
```diff
@@ -2,73 +2,72 @@
 
 Simulate any on-chain transaction, overriding the world state if needed.
 
-`simulation::SimulationEngine` can simulate an on-chain transaction and return tx output and state changes caused by
-that transaction.
+The `SimulationEngine` leverages `revm` to simulate on-chain transactions, returning both transaction output and state
+changes caused by the transaction. I also supports overriding account balances and storage slots.
```
tamaralipows, line 6:
```suggestion
changes caused by the transaction. It also supports overriding account balances and storage slots.
```

src/evm/Readme.md
```diff
@@ -2,73 +2,72 @@
...
 
-To do that, the engine needs access to the EVM state. This is provided by `database::SimulationDB`. This struct queries
-an ethereum node for needed data and caches it for a duration of 1 block. It also allows for setting up mocked accounts
-and overriding account balances and storage slots.
+To perform simulations, the engine requires access to the EVM state, provided by the `EngineDatabaseInterface` trait.
+Two
+implementations of this trait are available:
```
tamaralipows, line 10: Is it just looking weird because of the github UI? Probably
```suggestion
Two implementations of this trait are available:
```

src/evm/Readme.md
```diff
@@ -2,73 +2,72 @@
...
-Overrides set this way will take priority over accounts' permanent storage and querying from a node.
+- Non-Mocked Accounts (`mocked=false`):
+    - Missing storage slots are queried from the Ethereum node.
+- Mocked Accounts (`mocked=true`):
+    - No data is queried from the node.
+    - Missing storage slots default to 0 if they are not set in either permanent storage or simulation-time overrides.
 
-## See also
+Mocking is especially useful when overriding contract code, as the overridden code might rely on different storage slots
+than the on-chain version. Querying a node for such storage would not yield meaningful results.
```
tamaralipows, line 69: What does this actually mean, in non-diplomatic terms? haha. Can you give an example?

tamaralipows (commented):
Thanks! I have a few comments - in particular I'm worried about not linking to gitbook

src/evm/Readme.md
dianacarvalho1: oh this is weird..

src/evm/Readme.md
dianacarvalho1: I just re wrote what it said there 😂  but I think it's like you can mock the storage slots to be wtv you want. This is particularly useful if you are overriding the contract code to be different than on chain code, because the storage slots might be different

tamaralipows: I like what you just said a lot better

Readme.md
dianacarvalho1: we need to publish the docs before linking the docs though. but yes I agree

kayibal (commented):
Just FYI, the Readme under src/evm is like super old. Not sure we want to keep it. It certainly is a low level detail and should probably not make it into the GitBook.

src/evm/Readme.md
```diff
@@ -2,73 +2,71 @@
...
 
-Instead of querying account information or storage from a node, we can set it manually.
-
-This is done with `state.init_account()` method which takes the following arguments:
-
-- account address
-- `AccountInfo` instance
-- permanent storage mapping
-- `mocked` boolean
+Instead of relying on live queries for account data, you can manually configure account information using
```
louise-poole, line 35: What do you mean by 'live queries'? You mean querying the node?

src/evm/Readme.md
```diff
@@ -2,73 +2,71 @@
...
-
-An overriden account may be set as mocked or non-mocked.
-
-For non-mocked accounts (`mocked=false`), missing storage slots will be queried from a node.
+### Simulation-time overrides
 
-For mocked accounts (`mocked=true`), nothing will be queried from a node. Missing slots have value of 0: if a certain
-storage slot is needed during simulation, but has neither been set in the account's permanent storage (see the previous
-section) nor in the simulation-time overrides (see the next section), then the value of such slot will be 0.
+You can override the chain state only for the duration of a single simulation. This is done by
```
louise-poole, line 52: The word 'only' there means we don't support any other type of overrides, but we do... 'permanent_storage' above is a permanent override.
```suggestion
You can override the chain state for the duration of a single simulation. This is done by
```

src/evm/Readme.md
```diff
@@ -2,73 +2,71 @@
...
+setting `SimulationParameters.overrides` to a mapping of form `HashMap<Address, HashMap<U256, U256>>`.
```
louise-poole, line 53: What does this represent though? Would be nice to indicate that it's a [Address -> [slot -> value]] map I think.

src/evm/Readme.md
```diff
@@ -2,73 +2,71 @@
...
+permanent_storage: Option<HashMap<U256, U256>>,
+mocked: bool,
+```
 
-Permanent storage lets us set some storage slots. If a value of a certain storage slot is needed during simulation, the
-state will first check the permanent storage set here. Only if a slot is not set in the permanent storage, the state
-will query an Ethereum node (but see the next section).
+- `AccountInfo`: Contains core account details like balance, nonce, and code (for smart contract accounts).
+- `permanent_storage`: Allows setting predefined storage slots. When the simulation requires a specific storage slot,
+  the engine first checks this storage.
```
louise-poole, line 47: Hmmm... I wouldn't say 'first checks' this storage. Below you directly contradict this and say the simulation overrides are used before these permanent ones are. Maybe we should word it along the lines of: during simulations, the engine will prioritise the slots listed here above those provided in the db/by the data source?

src/evm/Readme.md
louise-poole: I agree - this can be worded better.

tamaralipows (approved):
What a beaut. Let's just remember to link gitbook eventually!

PR #98, opened 2024-12-13
docs: ProtocolStreamBuilder docs

tamaralipows:


src/evm/decoder.rs
```diff
@@ -50,6 +50,8 @@ pub(super) struct TychoStreamDecoder {
     inclusion_filters: HashMap<String, FilterFn>,
 }
 
+/// Decodes a `FeedMessage` into a `BlockUpdate` containing the updated states of protocol
+/// components for a specific block.
```
kayibal, line 54: Could you please convert this to have short heading and then more descriptive paragraph of text under it.

kayibal:
```suggestion
// Decodes incoming raw messages
//
// This struct decodes incoming messages of type `FeedMessage` and converts it into the `BlockUpdate`struct...
```

src/evm/decoder.rs
```diff
@@ -61,12 +63,20 @@ impl TychoStreamDecoder {
         }
     }
 
+    /// Sets the currently known tokens which will be considered during decoding.
+    ///
+    /// Protocol components containing tokens which are not included in this initial list, or
+    /// added when applying deltas, will not be decoded.
```
kayibal, line 69: This is important to mention on the struct docs as well..

src/evm/decoder.rs
```diff
@@ -61,12 +63,20 @@ impl TychoStreamDecoder {
...
         let mut guard = self.state.write().await;
         guard.tokens = tokens;
     }
 
-    // Method to register a decoder
+    /// Registers a decoder for a given exchange.
+    ///
+    /// Each of the native protocol states, along with the EVMPoolState, have their own decoders.
+    /// For example, to register a decoder for the `uniswap_v2` exchange, you must call
+    /// this function with `register_decoder::<UniswapV2State>("uniswap_v2")`.
```
kayibal, line 79: Would rather I would explain here that this associates a type with an exchanges id. It is required that the type implement `TryFromWithBlock`.

src/evm/decoder.rs
```diff
@@ -90,11 +100,22 @@ impl TychoStreamDecoder {
             .insert(exchange.to_string(), decoder);
     }
 
+    /// Registers a filter function for a given exchange.
+    ///
+    /// The filter function is used to determine whether a given component should be included in
+    /// the state updates. This can be useful when only wanting to retrieve state updates for
+    /// components with a TVL higher than a given amount, for example.
```
kayibal, line 107: No this is not true, this are client side pool filter functions. We can use them to express more customized conditions that are currently not supported by the server, e.g. to filter out pools with certain attributes because e.g. they are currently not fully supported - or to implement a blacklist etc.

src/evm/decoder.rs
```diff
@@ -90,11 +100,22 @@ impl TychoStreamDecoder {
...
         self.inclusion_filters
             .insert(exchange.to_string(), predicate);
     }
 
+    /// Decodes a `FeedMessage` into a `BlockUpdate` containing the updated states of protocol
+    /// components.
+    ///
+    /// The `BlockUpdate` will only contain updated protocol states for protocols components whose
+    /// protocols are registered, which pass the added inclusion filters, and whose token qualities
+    /// are above the minimum specified token quality.
```
kayibal, line 118: Imo should be mentioned first on struct level docs then mentioned briefly again here.

src/evm/stream.rs
```diff
@@ -21,6 +21,19 @@ pub struct ProtocolStreamBuilder {
...
+/// the block, such as the updated states of protocol components, which can in turn be used
+/// to obtain spot price information for a desired component and token pair.
+///
+/// # Returns
+/// A result containing a stream of decoded block updates, where each item is either:
+/// - `Ok(BlockUpdate)` if decoding succeeds.
+/// - `Err(StreamDecodeError)` if a decoding error occurs.
+///
+/// # Errors
+/// Returns a `StreamError` if the underlying stream builder fails to initialize.
```
kayibal, line 36: Please add a section here like `# Important` that briefly touches on how decoding works (e.g. if tokens are missing, the pool is skipped, etc.) You can do that by linking to the decoder struct docs.

src/evm/stream.rs
```diff
@@ -52,33 +68,42 @@ impl ProtocolStreamBuilder {
...
+    /// Disables TLS/ SSL for the connection, using http and ws protocols.
     pub fn no_tls(mut self, no_tls: bool) -> Self {
         self.stream_builder = self.stream_builder.no_tls(no_tls);
         self
     }
 
+    /// Sets the currently known tokens which to be considered during decoding.
+    ///
+    /// Protocol components containing tokens which are not included in this initial list, or
+    /// added when applying deltas, will not be decoded.
```
kayibal, line 106: This is so crucial information it should really be mentioned on the structs docs.

src/evm/decoder.rs
```diff
@@ -44,6 +44,21 @@ type RegistryFn =
     dyn Fn(ComponentWithState, Header, Arc<RwLock<DecoderState>>) -> DecodeFut + Send + Sync;
 type FilterFn = fn(&ComponentWithState) -> bool;
 
+/// A builder for creating protocol streams, decoding blocks, and managing protocol state updates.
```
kayibal, line 47: This is not a builder? Also please note that this is a private class (public only to the parent module).

src/evm/decoder.rs
```diff
@@ -44,6 +44,21 @@ type RegistryFn =
...
+/// qualities are above the minimum specified token quality.
+///
+/// ### Key Features:
+/// - Supports registering exchanges and their associated filters for specific protocol components.
+/// - Allows the addition of client-side filters for custom conditions.
+///
+/// **Note:** The builder uses the `TychoStreamBuilder` to fetch blocks and the `TychoStreamDecoder`
+/// to decode them. The tokens provided during configuration will be used for decoding, ensuring
+/// efficient handling of protocol components. Protocol components containing tokens which are not
+/// included in this initial list, or added when applying deltas, will not be decoded.
```
kayibal, line 61: Looks very chatgpt

src/evm/decoder.rs
```diff
@@ -52,6 +67,10 @@ pub(super) struct TychoStreamDecoder {
     inclusion_filters: HashMap<String, FilterFn>,
 }
 
+/// Decodes incoming raw messages.
+///
+/// This struct decodes incoming messages of type `FeedMessage` and converts it into the
+/// `BlockUpdate`struct
```
kayibal, line 73: Docs on a impl?

src/evm/stream.rs
```diff
@@ -18,6 +18,34 @@ use crate::{
...
 };
 
+/// Builds the protocol stream, providing a `BlockUpdate` for each block received.
+///
+/// Each `BlockUpdate` can then be used at a higher level to retrieve important information from
+/// the block, such as the updated states of protocol components, which can in turn be used
+/// to obtain spot price information for a desired component and token pair.
+///
+/// # Important
+/// Decoding is performed using the [`TychoStreamDecoder`](crate::evm::decoder::TychoStreamDecoder).
```
kayibal, line 28: I just realised that this class is private, so we can't link to it's docs :/ So let's move the corresponding docs from there to this struct, I think it is better to have it all here.

src/evm/decoder.rs
dianacarvalho1: you literally asked for this on the previous review 😂 see [here](https://github.com/propeller-heads/tycho-simulation/pull/98/files/58a469f0da2750e677bf8660a03f2e34219ee650#r1884988796). I'm removing it then

PR #100, opened 2024-12-14
chore: Group imports automatically

kayibal:


zizou0x (approved):

src/evm/account_storage.rs
```diff
@@ -229,11 +229,11 @@ impl AccountStorage {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use std::{collections::HashMap, error::Error, str::FromStr};
 
     use revm::primitives::{AccountInfo, Address, KECCAK_EMPTY, U256};
-    use std::{collections::HashMap, error::Error, str::FromStr};
 
+    use super::*;
```
dianacarvalho1, line 236: shouldn't this always be first? 😬

PR #101, opened 2024-12-16
feat: Update tycho dependencies tag and re export them

dianacarvalho1:


src/lib.rs
```diff
@@ -5,20 +5,14 @@
...
-//! The crate also provides optimization methods, such as golden
-//! section search, to find optimal amounts for a specific sequence.
 
 extern crate core;
 
 // Reexports
-pub use num_traits;
+pub use tycho_client;
+pub use tycho_core;
+pub use tycho_ethereum;
```
kayibal, line 14: Is this one required to be reexported?

dianacarvalho1: tycho_ethereum? huumm I'm not sure. If I'm being honest I don't even know what's in it 😅

dianacarvalho1: it's actually not even used in the simulation repo.. 🤷🏼‍♀️  removing it

kayibal (approved):
Ok looks a lot better now

PR #103, opened 2024-12-17
feat: Add tokens to ProtocolSim and use them for validations

dianacarvalho1:
In every get_amount and spot_prices we are now validating the tokens that are passed. This way, the user can't call these methods with random tokens and get a silently wrong result (no warning or error)

### DIFF
--- a/src/protocol/state.rs
+++ b/src/protocol/state.rs
@@ -25,12 +25,6 @@
 //! use tycho_simulation::models::Token;
 //! use tycho_simulation::evm::protocol::u256_num::u256_to_biguint;
 //!
-//! // Initialize the UniswapV2 state with token reserves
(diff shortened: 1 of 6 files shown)


src/protocol/state.rs
```diff
@@ -121,6 +122,23 @@ pub trait ProtocolSim: std::fmt::Debug + Send + Sync + 'static {
...
+            .iter()
+            .all(|address| self.token_addresses().contains(address))
+        {
+            return Err(SimulationError::InvalidInput(
+                "One or both token addresses are not the tokens of this pool".to_string(),
+                None,
+            ));
+        }
+        Ok(())
+    }
```
dianacarvalho1, line 140: is this ok to do? to have an implementation already here?

dianacarvalho1:
put on hold

PR #108, opened 2024-12-19
fix: Allow for no tls connections when getting tokens from Tycho

dianacarvalho1:
This is necessary to run in the cluster

### DIFF
--- a/.github/workflows/tests-and-lints-template.yaml
+++ b/.github/workflows/tests-and-lints-template.yaml
@@ -70,7 +70,7 @@ jobs:
         uses: taiki-e/install-action@nextest
 
       - name: Test
-        run: cargo nextest run --workspace --lib --all-targets --all-features
+        run: cargo nextest run --workspace --lib --all-targets --all-features && cargo test --doc
 
   lint:
     name: Code Lint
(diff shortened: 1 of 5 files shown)


louise-poole (approved):

src/utils.rs
```diff
@@ -36,8 +36,13 @@ pub fn hexstring_to_vec(hexstring: &str) -> Result<Vec<u8>, SimulationError> {
 }
 
 /// Loads all tokens from Tycho and returns them as a Hashmap of address->Token.
-pub async fn load_all_tokens(tycho_url: &str, auth_key: Option<&str>) -> HashMap<Bytes, Token> {
-    let rpc_url = format!("https://{tycho_url}");
+pub async fn load_all_tokens(
```
zizou0x, line 39: I wonder if it makes sense or if we shouldn't just give `tycho_url` with the prefix already there?

.github/workflows/tests-and-lints-template.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
         uses: taiki-e/install-action@nextest
 
       - name: Test
-        run: cargo nextest run --workspace --lib --all-targets --all-features
+        run: cargo nextest run --workspace --lib --all-targets --all-features && cargo test --doc
```
zizou0x, line 73: 💯

src/utils.rs
dianacarvalho1: I think no, because the tycho url is also used in more places and the prefixes are added there as well :/ or do you mean to add the prefix only once in a centralized place?

PR #111, opened 2024-12-19
feat(uniswap_v4): Implement `swap()` for `UniswapV4State`

zizou0x:
This PR implement the core swap function on `UniswapV4State`, it's almost equivalent to Uniswap V3, only the fees accounting change (we aren't taking protocol fees into account in our UniswapV3 implementation)

I also moved the remaining shared functions into the `utils/uniswap` module


src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -1,26 +1,51 @@
...
 
-impl UniswapV4ProtocolFees {
-    pub fn new(zero2one: i32, one2zero: i32) -> Self {
-        Self { zero2one, one2zero }
+impl UniswapV4Fees {
+    pub fn new(zero_for_one: u32, one_for_zero: u32, lp_fee: u32) -> Self {
+        Self { zero_for_one, one_for_zero, lp_fee }
+    }
+
+    fn calculate_swap_fees_pips(&self, zero_for_one: bool) -> u32 {
```
zizou0x, line 46: This can be considered as shallow, and we could probably have this directly in the functions body. But I decided to do it like this to stay consistent with how it's done in their smart contract

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -43,6 +67,200 @@ impl UniswapV4State {
...
+            next_tick = next_tick.clamp(MIN_TICK, MAX_TICK);
+
+            let sqrt_price_next = get_sqrt_ratio_at_tick(next_tick)?;
+            let (sqrt_price, amount_in, amount_out, fee_amount) = swap_math::compute_swap_step(
+                state.sqrt_price,
+                UniswapV4State::get_sqrt_ratio_target(sqrt_price_next, price_limit, zero_for_one),
+                state.liquidity,
+                state.amount_remaining,
+                self.fees
+                    .calculate_swap_fees_pips(zero_for_one),
```
louise-poole, line 145: So basically this line is the only difference with usv3 right now? 👀

louise-poole (approved):
Looks good!

src/evm/protocol/uniswap_v4/state.rs
zizou0x: It's not even really a difference because it's there in V3, we just don't have it implemented because they never decided to enable protocol fees.

I'd say the biggest difference I've seen do far is that the `fees` and `tick_spacing` can be anything and there is no map `fee => tick_spacing` anymore. I'm looking forward to test all this with actual on chain state to confirm though

src/evm/protocol/uniswap_v4/state.rs
tamaralipows, line 46: What are pips?

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -43,6 +67,200 @@ impl UniswapV4State {
...
-        UniswapV4State { liquidity, sqrt_price, lp_fee, protocol_fees, tick, ticks: tick_list }
+        UniswapV4State { liquidity, sqrt_price, fees, tick, ticks: tick_list }
+    }
+
+    fn swap(
+        &self,
+        zero_for_one: bool,
+        amount_specified: I256,
+        sqrt_price_limit: Option<U256>,
+    ) -> Result<SwapResults, SimulationError> {
```
tamaralipows, line 78: Sorry I'm not very familiar with USV4.

But why are we implementing a `swap` method with `SwapResults` instead of a `get_amount_out` like all the other pool states? If there is a good reason perhaps we should document that in the code?

dianacarvalho1: this is a UniswapV4 specific method. then the implementation of `ProtocolSim` for uni v4 will have the `get_amount_out` that will eventually call this method!

src/evm/protocol/uniswap_v4/state.rs
zizou0x: It's the unit they use for fees, basically `1 pip = 0.01%`

Not sure what's the diff between pips and bips though, I've seen both in their contract

src/evm/protocol/uniswap_v4/state.rs
zizou0x: `get_amount_out` will be implemented as part of the `ProtocolSim` trait implementation. This function is just an internal function that will be used in `get_amount_out`.

I suppose we could directly have everything inside `get_amount_out`, but maybe it would be too much for a single function? 

For now I think it's alright like this because it's consistent with `UniswapV3`

tamaralipows: Ok ok nice! You can keep it as is - no worries

src/evm/protocol/uniswap_v4/state.rs
tamaralipows: ah ok! I've heard of bips, never pips. cool

tvinagre:
```
    // The protocol fee is taken from the input amount first and then the LP fee is taken from the remaining
    // The swap fee is capped at 100%
    // Equivalent to protocolFee + lpFee(1_000_000 - protocolFee) / 1_000_000 (rounded up)
    /// @dev here `self` is just a single direction's protocol fee, not a packed type of 2 protocol fees
    function calculateSwapFee(uint16 self, uint24 lpFee) internal pure returns (uint24 swapFee) {
        // protocolFee + lpFee - (protocolFee * lpFee / 1_000_000)
        assembly ("memory-safe") {
            self := and(self, 0xfff)
            lpFee := and(lpFee, 0xffffff)
            let numerator := mul(self, lpFee)
            swapFee := sub(add(self, lpFee), div(numerator, PIPS_DENOMINATOR))
        }
    }
```

Shouldn't you do the division or am I missing smth?

tvinagre (commented):
Looking good - I suggest you copy over some of the tests that they use on the contract - I've used them to validate that my swapMath was good on Substreams side

src/evm/protocol/uniswap_v4/state.rs
zizou0x: From what I saw we use pips in our maths, I'm not 100% sure how the substreams gives us things yet though, this will need to be confirmed when testing

PR #113, opened 2024-12-19
feat(uniswap_v4): implement `ProtocolSim` except `delta_transition`

zizou0x:
`delta_transition` will be implemented in a future PR because it depends on the Substreams.

### DIFF
--- a/.github/workflows/tests-and-lints-template.yaml
+++ b/.github/workflows/tests-and-lints-template.yaml
@@ -52,9 +52,9 @@ jobs:
 
       - name: Setup git to use https
         run: |
-            git config --global credential.helper store
-            echo "https://${{ steps.generate-token.outputs.token }}@github.com" > ~/.git-credentials
-            git config --global url."https://x-access-token:${{ steps.generate-token.outputs.token }}@github.com".insteadOf ssh://github.com
+          git config --global credential.helper store
+          echo "https://${{ steps.generate-token.outputs.token }}@github.com" > ~/.git-credentials
+          git config --global url."https://x-access-token:${{ steps.generate-token.outputs.token }}@github.com".insteadOf ssh://github.com
 
       - name: Setup toolchain
         uses: dtolnay/rust-toolchain@v1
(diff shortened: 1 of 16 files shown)


tvinagre (approved):

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -217,50 +227,92 @@ impl UniswapV4State {
     }
 }
 
-#[allow(unused_variables)] //TODO: remove when implemented
 impl ProtocolSim for UniswapV4State {
+    // Not possible to implement correctly with the current interface because we need to know the
+    // swap direction.
```
louise-poole, line 232: Are we going to not implement it at all? Where is it used? If needed we can return some sort of best guess - like a minimum fee?

louise-poole (approved):

src/evm/protocol/uniswap_v4/state.rs
zizou0x: It's not used AFAIK, this one isn't implemented for VMs also. We need to find a way to make this interface better, but in the scope of this epic it's not planned to be implemented no. 

The minimum fee was also suggested by Alan, but I find it unexpected from a client POV so I decided to leave it todo for now. 
We can easily come back to this later when they launch, for example if they don't enable any protocol fee (similar to uniswapv3) then we can return lp fee there.

dianacarvalho1: yes, fees are not supported in this simulation release.. we need to come back to it at some point

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -217,50 +227,92 @@ impl UniswapV4State {
...
+        token_in: &Token,
+        token_out: &Token,
     ) -> Result<GetAmountOutResult, SimulationError> {
-        todo!()
+        let zero_for_one = token_in < token_out;
+        let amount_specified = I256::checked_from_sign_and_abs(
+            Sign::Positive,
+            U256::from_be_slice(&amount_in.to_bytes_be()),
+        )
+        .unwrap();
```
dianacarvalho1, line 261: can you use expect instead of unwrap here? 👀

dianacarvalho1 (approved):
Thank you @zizou0x ! tiny tiny comment only!

src/evm/protocol/uniswap_v4/state.rs
louise-poole: So proud of you, @dianacarvalho1 ❤️ 🤣

dianacarvalho1: I've levelled up in my Rust track!! ➕

PR #114, opened 2024-12-19
fix: Load bytecode from files with include_bytes!

dianacarvalho1:
This way the bytecode is loaded at compile time and not at run time, this way we won't have problems with weird paths  (I hope)

### DIFF
--- a/src/evm/protocol/vm/constants.rs
+++ b/src/evm/protocol/vm/constants.rs
@@ -1,10 +1,25 @@
 use alloy_primitives::{Address, U256};
 use lazy_static::lazy_static;
 
+use crate::protocol::errors::SimulationError;
+
 lazy_static! {
     pub static ref EXTERNAL_ACCOUNT: Address = Address::from_slice(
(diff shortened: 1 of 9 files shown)


src/evm/protocol/vm/constants.rs
```diff
@@ -1,10 +1,25 @@
...
+pub const CURVE: &[u8] = include_bytes!("assets/CurveSwapAdapter.evm.runtime");
+pub fn get_adapter_file(protocol: &str) -> Result<&'static [u8], SimulationError> {
+    match protocol {
+        "balancer_v2" => Ok(BALANCER_V2),
+        "curve" => Ok(CURVE),
+        _ => {
+            Err(SimulationError::FatalError(format!("Adapter for protocol {} not found", protocol)))
+        }
+    }
+}
```
dianacarvalho1, line 25: I know that if this stays I need to update the docs 🫠

kayibal (approved):

PR #117, opened 2024-12-20
feat(uniswap_v4): add filter for pools with hooks

zizou0x:


src/evm/protocol/filters.rs
```diff
@@ -123,3 +123,22 @@ pub fn curve_pool_filter(component: &ComponentWithState) -> bool {
...
+        .static_attributes
+        .get("hooks")
+    {
+        if !hooks.to_vec().iter().all(|&b| b == 0) {
+            debug!("Filtering out UniswapV4 pool {} because it has hooks", component.component.id);
+            return false;
+        }
+    }
+    debug!(
+        "UniswapV4 pool with static attributes {:?} will not be filtered out.",
```
zizou0x, line 140: I think I wanna change this, why do we do this in others filters?

src/evm/protocol/filters.rs
```diff
@@ -123,3 +123,22 @@ pub fn curve_pool_filter(component: &ComponentWithState) -> bool {
...
 }
+
+/// Filters out pool that have hooks in Uniswap V4
+pub fn uniswap_v4_pool_with_hook_filter(component: &ComponentWithState) -> bool {
+    if let Some(hooks) = component
+        .component
+        .static_attributes
+        .get("hooks")
+    {
+        if !hooks.to_vec().iter().all(|&b| b == 0) {
```
tvinagre, line 134: Comparing to a vec made from the address 0 would be cleaner IMO, but not the biggest deal. Pls just add a comment maybe saying the address 0 == no hook

tvinagre (approved):

src/evm/protocol/filters.rs
tamaralipows: If you based this off the Balancer filter it was purely for my debugging because I couldn't figure out why some pools were not getting filtered out. Feel free to remove this!

dianacarvalho1: yeeess feel free to delete these logs! I had already deleted them but then got brought back from the dead in some bad merge

PR #124, opened 2025-01-09
feat(price_printer): dynamically adapt the quote amount to the new decimals when switching row

zizou0x:
Previously, the quote amount remained unchanged when switching the target components. This caused issues, such as trading excessively large amounts when transitioning from a token with 18 decimals to one with 6 decimals (as in token).

This update ensures that the quote amount is correctly scaled to match the decimals of the new token when changing the target pool.

### DIFF
--- a/examples/price_printer/ui.rs
+++ b/examples/price_printer/ui.rs
@@ -98,70 +98,36 @@ impl App {
             items: data_vec,
         }
     }
-    pub fn next_row(&mut self) {
-        let mut current_decimals = None;
-        if let Some(idx) = self.state.selected() {
-            let comp = &self.items[idx].component;
-            current_decimals = if self.zero2one {
-                Some(comp.tokens[0].decimals)
-            } else {
-                Some(comp.tokens[1].decimals)
-            };
-        }
 
-        let i = match self.state.selected() {
+    pub fn move_row(&mut self, direction: isize) {
+        // Get current decimals, if any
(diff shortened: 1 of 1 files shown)


examples/price_printer/ui.rs
```diff
@@ -113,9 +123,28 @@ impl App {
...
+
+        if let (Some(prev_decimals), Some(idx)) = (current_decimals, self.state.selected()) {
+            let comp = &self.items[idx].component;
+            let decimals = comp.tokens[if self.zero2one { 0 } else { 1 }].decimals;
+            if decimals >= prev_decimals {
+                self.quote_amount *= BigUint::from(10u64).pow((decimals - prev_decimals) as u32);
+            } else {
+                self.quote_amount /= BigUint::from(10u64).pow((prev_decimals - decimals) as u32);
+            }
+        }
```
dianacarvalho1, line 135: I would just suggest to put this in a function called `scale_quote_amount` or something. But this is not a big duplication, so I won't insist!

dianacarvalho1 (approved):
Thank you @zizou0x 🙏🏼 Only have a small suggestion!

examples/price_printer/ui.rs
zizou0x: Addressed in https://github.com/propeller-heads/tycho-simulation/pull/124/commits/07e63d3348ed60f83a2d2a2b1c24df3246a4cda6

PR #125, opened 2025-01-20
feat(balancer-v3): add a specific overwrite case for BalancerV3 vault

zizou0x:
This is a hack and it is needed because of an assertion in BalancerV3 Vault contract that can't pass with how we currently overwrite tokens. There is a plan to fix this properly in the future


tvinagre (approved):

tvinagre:
Hoping to revert this soon!

PR #130, opened 2025-01-28
feat: add uniswap_v4 to examples

tvinagre:


examples/price_printer/main.rs
```diff
@@ -1,8 +1,7 @@
+extern crate tycho_simulation;
 mod ui;
 pub mod utils;
 
-extern crate tycho_simulation;
-
```
louise-poole, line 5: Why this change? It's nicer if internal mods and external imports are separate....

tvinagre: blame cargo fmt 😢

louise-poole (approved):

PR #133, opened 2025-02-04
feat: better support switching chains

louise-poole:
Includes:
- updating the tycho-indexer version used to one that supports Base chain
- removing unnecessary re-implementations of tycho core enums
- add chain argument to load_all_tokens instead of hardcoding it to Ethereum

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7071,8 +7071,8 @@ dependencies = [
 
 [[package]]
 name = "tycho-client"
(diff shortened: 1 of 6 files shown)


dianacarvalho1 (approved):
oh uau! amazing that these were the only changes needed 👏🏼 
Thank you @louise-poole 🙏🏼

tvinagre (approved):

PR #134, opened 2025-02-04
feat: Refactor quickstart and add execution

dianacarvalho1:
Refactor: Instead of looping through the top 10 pools and calculating spot prices and amount out for each block, it will return the best amount out for a token pair. The trade tokens can be defined through CLI parameters. 

Add execution to quickstart. The user can choose to simulate, execute or skip the swap. The output looks like this
````
Looking for the best swap for 1 WETH -> USDC
==================== Received block 21937897 ====================
The best swap (out of 6 possible pools) is:
protocol: "uniswap_v3"
id: "0x88e6a0c2ddd26feeb64f039a2c41296fcb3f5640"
swap: 1000000000000000000 WETH -> 2351701017 USDC
Do you want to simulate, execute or skip this swap?
Please be aware that the market might move while you make your decision. Which might lead to a revert if you've set a min amount out or slippage.
(simulate/execute/skip): 
````
if they choose `simulate`
````
(simulate/execute/skip): simulate
Simulating by performing an approval (for permit2) and a swap transaction...
Simulated Block 21937944:
  Transaction 1: Status: true, Gas Used: 45992
  Transaction 2: Status: false, Gas Used: 66574
````
if they choose `execute`
````
(simulate/execute/skip): execute
Executing by performing an approval (for permit2) and a swap transaction...
Approval transaction sent with hash: 0x267fc0027c861bd740d9752fde2d3b44c6db148c3da7c4fbab082feeb484bb89 and status: true
Swap transaction sent with hash: 0x330458ba7805a9647fa08f3bf4c7634f8975f1ea7d2b3e3e135916ddbcc90926 and status: true
````

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ lazy_static = "1.4.0"
 # Tycho dependencies
 tycho-core = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-core", tag = "0.56.5" }
 tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-client", tag = "0.56.5" }
-tycho-execution = { git = "https://github.com/propeller-heads/tycho-execution.git", package = "tycho-execution", rev = "58749bd65b697673aff94f75954dff5504182041", features = ["evm"] }
+tycho-execution = { git = "https://github.com/propeller-heads/tycho-execution.git", package = "tycho-execution", rev = "1e2c628ba969594709912157b0b8df816684b95d", features = ["evm"] }
 
 # EVM dependencies
 foundry-config = { git = "https://github.com/foundry-rs/foundry", rev = "57bb12e", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ lazy_static = "1.4.0"
 # Tycho dependencies
 tycho-core = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-core", tag = "0.56.5" }
 tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-client", tag = "0.56.5" }
+tycho-execution = { git = "https://github.com/propeller-heads/tycho-execution.git", package = "tycho-execution", rev = "58749bd65b697673aff94f75954dff5504182041", features = ["evm"] }
 
 # EVM dependencies
 foundry-config = { git = "https://github.com/foundry-rs/foundry", rev = "57bb12e", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ alloy-primitives = { version = "0.8.9", features = [
     "map-foldhash",
 ] }
 alloy-sol-types = { version = "0.8.14" }
-alloy = { version = "0.5.4", features = ["providers"] }
+alloy = { version = "0.5.4", features = ["providers", "signer-local", "rpc-types-eth"] }
 revm = { version = "17.1.0", features = ["ethersdb", "serde"], optional = true }
 revm-inspectors = { version = "0.10", features = ["serde"], optional = true }
 num-bigint = "0.4.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ lazy_static = "1.4.0"
 # Tycho dependencies
 tycho-core = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-core", tag = "0.56.5" }
 tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-client", tag = "0.56.5" }
-tycho-execution = { git = "https://github.com/propeller-heads/tycho-execution.git", package = "tycho-execution", features = ["evm"], rev = "05a1843f9c3f916d4af8fbf67cccf3b3b94aeab8" }
+tycho-execution = { git = "https://github.com/propeller-heads/tycho-execution.git", package = "tycho-execution", features = ["evm"], tag = "0.54.0" }
 
 # EVM dependencies
 foundry-config = { git = "https://github.com/foundry-rs/foundry", rev = "57bb12e", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ lazy_static = "1.4.0"
 # Tycho dependencies
 tycho-core = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-core", tag = "0.56.5" }
 tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-client", tag = "0.56.5" }
-tycho-execution = { git = "https://github.com/propeller-heads/tycho-execution.git", package = "tycho-execution", features = ["evm"], tag = "0.51.0" }
+tycho-execution = { git = "https://github.com/propeller-heads/tycho-execution.git", package = "tycho-execution", features = ["evm"], rev = "5aec1c08117c249ebe68421367ad6a68636fca9b" }
 
 # EVM dependencies
 foundry-config = { git = "https://github.com/foundry-rs/foundry", rev = "57bb12e", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ lazy_static = "1.4.0"
 # Tycho dependencies
 tycho-core = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-core", tag = "0.56.5" }
 tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-client", tag = "0.56.5" }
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: examples/quickstart/main.rs [hunk 1]
```diff
@@ -61,43 +101,70 @@ async fn main() {
...
+    encode_best_amount_out(amounts_out);
+}
+
+fn encode_best_amount_out(amounts_out: &HashMap<String, BigUint>) {
+    if let Some((key, amount_out)) = amounts_out
+        .iter()
+        .max_by_key(|(_, value)| value.to_owned())
+    {
+        println!(
+            "Pool with the highest amount out: {} with {} (out of {} results)",
```
dianacarvalho1, line 161: I'm getting very little pools here.. I will investigate tomorrow 🔍

dianacarvalho1: no, it looks good. AG

Cargo.toml
```diff
@@ -40,6 +40,7 @@ lazy_static = "1.4.0"
 # Tycho dependencies
 tycho-core = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-core", tag = "0.55.2" }
 tycho-client = { git = "https://github.com/propeller-heads/tycho-indexer.git", package = "tycho-client", tag = "0.55.2" }
+tycho-execution = { git = "https://github.com/propeller-heads/tycho-execution.git", package = "tycho-execution", rev = "58749bd65b697673aff94f75954dff5504182041", features = ["evm"] }
```
dianacarvalho1, line 43: Once we merge [this PR](https://github.com/propeller-heads/tycho-execution/pull/57), I will point this to a version

examples/quickstart/main.rs
```diff
@@ -33,16 +62,37 @@ async fn main() {
...
+    println!(
+        "Looking for the best swap for {} {} -> {}",
+        cli.sell_amount, sell_token.symbol, buy_token.symbol
+    );
+    let mut pairs: HashMap<String, ProtocolComponent> = HashMap::new();
+    let mut amounts_out: HashMap<String, BigUint> = HashMap::new();
 
     let mut protocol_stream = ProtocolStreamBuilder::new(&tycho_url, Chain::Ethereum)
         .exchange::<UniswapV2State>("uniswap_v2", tvl_filter.clone(), None)
+        .exchange::<UniswapV3State>("uniswap_v3", tvl_filter.clone(), None)
```
dianacarvalho1, line 95: @tamaralipows this is not working for uniswap v3 yet. There is a problem when decoding the fee 😕 I thought it was only the key that was wrong (and changed it [here](https://github.com/propeller-heads/tycho-execution/pull/57/commits/5eaf5315716dc481a00248c3b21f4a75a9c23b6b)) but it seems we are decoding wrong too. If you have some time, can you have a look please? Just run the quickstart in this branch. Right before it fails to encode you can print the whole component to see how the fee really looks like!

tamaralipows: Just saw this now - I will start tomorrow by looking into it then!

tamaralipows (approved):
Thank you! Already updating gitbook with this info

examples/quickstart/Readme.md
```diff
@@ -14,5 +12,13 @@ export RPC_URL=<your-eth-rpc-url>
 cargo run --release --example quickstart
 ```
 
+By default, the example will trade 1 WETH -> USDC. If you want a different trade you can do:
+
+```bash
+cargo run --release --example quickstart -- --sell-token "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48" --buy-token "0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599" --sell-amount 10000
```
tamaralipows, line 18: Ah quick question, just noticed this now, what is with this `--` after quickstart? Haven't tried to run this yet

examples/quickstart/main.rs
```diff
@@ -1,26 +1,55 @@
...
+#[derive(Parser)]
+struct Cli {
+    #[arg(short, long, default_value = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2")]
+    sell_token: String,
+    #[arg(short, long, default_value = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48")]
+    buy_token: String,
+    #[arg(short, long, default_value_t = 1)]
+    sell_amount: u32,
+    /// The tvl threshold to filter the graph by
+    #[arg(short, long, default_value_t = 100.0)]
```
royvardhan, line 49: Is 100 representing a value in usd?

royvardhan (approved):
🙌

examples/quickstart/Readme.md
dianacarvalho1: it's necessary for clap to recognise whatever comes next as the CLI args

examples/quickstart/main.rs
dianacarvalho1: no no, it's ETH

examples/quickstart/main.rs
```diff
@@ -150,22 +188,67 @@ fn get_best_swap(
...
+        Some(key.to_string())
     } else {
         println!("There aren't pools with the tokens we are looking for");
+        None
     }
 }
+
+fn encode(
+    encoder: EVMTychoEncoder<EVMStrategyEncoderRegistry>,
+    pairs: &HashMap<String, ProtocolComponent>,
```
louise-poole, line 210: Why pass the whole HashMap of pairs? This would be a simpler interface (and probably better practice?) to pass only the ProtocolComponent that will be used.
It means you can also reduce the number of args here... instead of both `pairs` and `best_pool` you can pass only `pool: ProtocolComponent`?

examples/quickstart/main.rs
```diff
@@ -100,39 +111,66 @@ async fn main() {
...
     }
     if message.states.is_empty() {
         println!("No pools of interest were updated this block. The best swap is the previous one");
-        return
+        return None;
     }
     for (id, state) in message.states.iter() {
-        if let Some(tokens) = pairs.get(id) {
+        if let Some(component) = pairs.get(id) {
+            let tokens = component.tokens.clone();
```
louise-poole, line 173: I might be wrong here... but does this need to be cloned? It looks like you use references to the tokens from here onwards so don't need to clone, no?

examples/quickstart/main.rs
```diff
@@ -67,45 +113,144 @@ async fn main() {
         .await
         .expect("Failed building protocol stream");
 
+    // execution setup
+    let swapper_pk =
+        "0x123456789abcdef123456789abcdef123456789abcdef123456789abcdef1234".to_string();
+    let user_address = Bytes::from_str("0xcd09f75E2BF2A4d11F3AB23f1389FcC1621c0cc2")
```
royvardhan, line 119: Why does it use the WETH address as the user address?

royvardhan:
[This](https://github.com/propeller-heads/tycho-simulation/blob/bea756f4b35c1f714ef30716a9018526ad8295c5/examples/quickstart/Readme.md?plain=1#L11) should be ETH_RPC_URL

examples/quickstart/main.rs
```diff
@@ -45,20 +91,41 @@ async fn main() {
...
+        // .exchange::<EVMPoolState<PreCachedDB>>(
+        //     "vm:balancer_v2",
+        //     tvl_filter.clone(),
+        //     Some(balancer_pool_filter),
+        // )
+        // .exchange::<UniswapV4State>(
+        //     "uniswap_v4",
+        //     tvl_filter.clone(),
+        //     Some(uniswap_v4_pool_with_hook_filter),
+        // )
```
dianacarvalho1, line 128: ups, I will bring them back later

tamaralipows (approved):
Thank you so much for figuring this out! And the code is very clean., I'm happy

examples/quickstart/main.rs
```diff
@@ -190,7 +190,9 @@ async fn main() {
                 user_address.clone(),
             );
 
-            print!("Do you want to simulate, execute or skip this swap? (simulate/execute/skip): ");
+            println!("Do you want to simulate, execute or skip this swap?");
+            println!("Please be aware that the market might move while you make your decision.");
```
tamaralipows, line 194: "...leading to a revert if you've set a min amount out or slippage."

Unless they will see the revert message nicely if it reverts?

dianacarvalho1: no, they don't see a nice message 🙃

examples/quickstart/main.rs
```diff
@@ -67,45 +138,325 @@ async fn main() {
...
+                amount_in.clone(),
+                Bytes::from(wallet.address().to_vec()),
+            );
+
+            if cli.swapper_pk == FAKE_PK {
+                println!("Signer private key was not provided. Skipping simulation/execution...");
+                continue
+            }
+            println!("Do you want to simulate, execute or skip this swap?");
+            println!("Please be aware that the market might move while you make your decision. Which might lead to a revert if you've set a min amount out or slippage.");
```
tamaralipows, line 195: sry nitpick
```suggestion
            println!("Please be aware that the market might move while you make your decision, which might lead to a revert if you've set a min amount out or slippage.");
```

PR #135, opened 2025-02-05
feat: Extend ProtocolComponent to have the attributes of the core model

dianacarvalho1:
Also implement From<ProtocolComponent> for tycho_core::dto::ProtocolComponent. This will be useful for tycho-execution

Deprecate address in ProtocolComponent. It should be id instead

(this was decided [here](https://propellerswap.slack.com/archives/C07H7S15LUV/p1738775735122579?thread_ts=1738687888.096819&cid=C07H7S15LUV))

### DIFF
--- a/src/protocol/models.rs
+++ b/src/protocol/models.rs
@@ -24,11 +24,12 @@
 //! It's worth emphasizing that although the term "pair" used in this
 //! module refers to a trading pair, it does not necessarily imply two
 //! tokens only. Some pairs might have more than two tokens.
-use std::{collections::HashMap, future::Future};
+use std::{collections::HashMap, default::Default, future::Future};
 
+use chrono::NaiveDateTime;
 use num_bigint::BigUint;
 use tycho_client::feed::Header;
-use tycho_core::Bytes;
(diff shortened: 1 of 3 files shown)


src/protocol/models.rs
```diff
@@ -41,14 +42,57 @@ use crate::models::Token;
...
+            id: id.clone(),
+            address: id,
+            tokens,
+            protocol_system: core_model.protocol_system,
+            protocol_type_name: core_model.protocol_type_name,
+            chain: core_model.chain,
+            contract_ids: core_model.contract_ids,
+            static_attributes: core_model.static_attributes,
+            creation_tx: core_model.creation_tx,
+            created_at: core_model.created_at,
```
tamaralipows, line 73: Do we need this? Or are we adding it just in case

tamaralipows (approved):
Thank you! Looks great, just a very minor question

royvardhan (approved):

src/protocol/models.rs
dianacarvalho1: adding just in case.. it's not a lot of info, so I think it's fine to have it!

PR #137, opened 2025-02-06
feat: add account balance overrides 

louise-poole:
For some protocols the balances used by a component is stored on a separate contract. To best cater for such components we now support using account_balances where the balance for relevant contracts is monitored and overwritten during simulation.

Decisions:
- it was decided to work towards deprecating the use 'balance_owners' and instead encourage using account balances. For backwards compatibility, support for protocols using balance_owner has been maintained.

### DIFF
--- a/src/evm/decoder.rs
+++ b/src/evm/decoder.rs
@@ -113,7 +113,7 @@ impl TychoStreamDecoder {
         let decoder = Box::new(
             move |component: ComponentWithState,
                   header: Header,
-                  account_balances: HashMap<Bytes, HashMap<Bytes, Bytes>>,
+                  account_balances: AccountBalances,
                   state: Arc<RwLock<DecoderState>>| {
                 Box::pin(async move {
                     let guard = state.read().await;
--- a/src/evm/decoder.rs
+++ b/src/evm/decoder.rs
@@ -560,7 +560,6 @@ mod tests {
     }
 
     fn load_test_msg(name: &str) -> FeedMessage {
-        dbg!(&name);
         let project_root = env!("CARGO_MANIFEST_DIR");
         let asset_path =
             Path::new(project_root).join(format!("tests/assets/decoder/{}.json", name));
--- a/src/evm/decoder.rs
+++ b/src/evm/decoder.rs
@@ -8,7 +8,7 @@ use std::{
 
 use alloy_primitives::Address;
 use thiserror::Error;
-use tokio::sync::RwLock;
+use tokio::sync::{RwLock, RwLockReadGuard};
 use tracing::{debug, error, info, warn};
 use tycho_client::feed::{synchronizer::ComponentWithState, FeedMessage, Header};
 use tycho_core::{dto::ProtocolStateDelta, Bytes};
@@ -354,6 +354,7 @@ impl TychoStreamDecoder {
 
             // PROCESS DELTAS
             if let Some(deltas) = protocol_msg.deltas.clone() {
+                // Update engine with account changes
                 let account_update_by_address: HashMap<Address, AccountUpdate> = deltas
                     .account_updates
                     .clone()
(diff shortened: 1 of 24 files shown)


src/evm/decoder.rs
```diff
@@ -279,8 +293,16 @@ impl TychoStreamDecoder {
...
+                    match state_decode_f(
+                        snapshot,
+                        block.clone(),
+                        account_balances.clone(),
+                        self.state.clone(),
+                    )
+                    .await
+                    {
                         Ok(state) => {
+                            // TODO: Handle account balances
```
zizou0x, line 305: What needs to be done here?

zizou0x (commented):
Looks neat! Just one question about a TODO

src/evm/decoder.rs
louise-poole: Nothing.... was left over when I was running through an initial design idea. 👀

src/evm/decoder.rs
```diff
@@ -349,9 +369,35 @@ impl TychoStreamDecoder {
                 .await;
                 info!("Engine updated");
 
-                // update states related to contracts with account deltas
+                // Collect all balance changes this block
+                let all_balances = Balances {
```
louise-poole, line 373: Currently with this approach we only update the balances on pools that experience either a contract change or state attribute change. Will we ever have the case where the only update from a component is isolated balance updates? If so, we need to add logic to update those too.

zizou0x: To be honest I don't see a case where it would happen now but I also don't feel safe to make the assumption that this will never happen

Would it be a big effort to cover this case?

src/evm/decoder.rs
```diff
@@ -500,6 +552,7 @@ mod tests {
     }
 
     fn load_test_msg(name: &str) -> FeedMessage {
+        dbg!(&name);
```
zizou0x, line 555: Shouldn't be commited

louise-poole: Ooops - pretend you never saw it...

zizou0x:
I noticed a lot of missing/not updated docs, would be nice to double check that we keep everything up to date before merging

src/evm/decoder.rs
louise-poole: I'd say medium effort. We could add a comment for the assumption and leave it as is for now if we're pressed for time.
To deal with this case we'd need to filter out the balances we've already applied and then call `delta_transition` for the pools that need their balances still updated. It also raises the question: could we expect there to be deltas for pools we've just decoded snapshots for? If so, we cannot use 'updated_pools' to filter out pools that already had their balances updated since it contains both snapshot (still to update) and delta updated pools.

src/evm/protocol/vm/tycho_decoder.rs
```diff
@@ -48,23 +48,42 @@ impl TryFromWithBlock<ComponentWithState> for EVMPoolState<PreCachedDB> {
     async fn try_from_with_block(
         snapshot: ComponentWithState,
         block: Header,
+        account_balances: &HashMap<Bytes, HashMap<Bytes, Bytes>>,
```
louise-poole, line 51: Idea: We could refactor our try_from_with_block to decode whole `Snapshot`s instead of `ComponentWithState` (as is done in the python version). Then we'd have the account balances inside the snapshot and don't need to collect and pass it extra like this.

src/evm/decoder.rs
kayibal: Are these the balances for this component or literally all balances?

src/evm/protocol/vm/state.rs
```diff
@@ -46,21 +47,25 @@ where
     pub tokens: Vec<Bytes>,
     /// The current block, will be used to set vm context
     block: BlockHeader,
-    /// The pool's token balances
+    /// The pool's token balances. This has been deprecated in favor of `contract_balances`.
+    #[deprecated(note = "Use contract_balances instead")]
```
kayibal, line 51: What, why no more `component_balances`? I don't think we should deprecate this... IMO component balances are still valid as a concept since they describe tvl. But maybe tvl is not required here for simulation and therefore it is removed? Hmm keeping it may be confusing then... lmk if this was what you were thinking here.

kayibal: Although if we deprecate this, then we always need to query contract balances and component balances... instead of just querying component balances when contract <-> component are 1:1.

src/evm/protocol/vm/state_builder.rs
```diff
@@ -221,26 +230,53 @@ where
...
-            })?,
-        ))
+
+        let adapter_contract = self.adapter_contract.ok_or_else(|| {
+            SimulationError::FatalError(
+                "Failed to get build engine: Adapter contract not initialized".to_string(),
+            )
+        })?;
+
+        if !self.contract_balances.is_empty() {
```
kayibal, line 240: Not sure I like this, especially if we keep supporting both we may have other options here?

src/evm/protocol/vm/state.rs
louise-poole: This was a glitch in thinking on my side - I meant to only deprecate the balance owner... not component balances entirely.

src/evm/decoder.rs
louise-poole: All balances. When we update the pool, at this level we do not know which contracts that pool uses. So we pass all contract balance updates and on the pool itself it can filter out for the appropriate balances.
We have a contract:pools map on this StreamDecoder (used to collect a set of pools to update), but the reverse map doesn't exist.

src/evm/protocol/vm/tycho_decoder.rs
tvinagre: Yes, that'd be the cleaner way. Can we do this in a separate PR?

louise-poole: But it means we pass a lot more data in the interface: The full snapshot has the full vm_storage with it which include storage slots, contract code etc.

src/evm/decoder.rs
louise-poole: Yes - if the vault isn't tracking its own reserves or what amounts are allocated to which pools, then it could potentially have a balance update without simultaneous account update.

tvinagre (approved):
🔥

PR #145, opened 2025-02-11
fix: decoder triggers state updates on related account changes

louise-poole:
Issue: previously when an account was updated, we reflected the changes on the simulation engine, but never triggered the related pool states to update. Therefore their spot prices etc were never recalculated on updates.

Solution: manage a component:pool ids map to keep track of which contracts affect which pool states. On a contract change, we then trigger state updates on all linked pool states. We skip adding pools to this map if the pool is configured for 'manual_updates' (their state updates are managed with update trigger state attribute changes).

### DIFF
--- a/src/evm/decoder.rs
+++ b/src/evm/decoder.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{hash_map::Entry, HashMap},
+    collections::{hash_map::Entry, HashMap, HashSet},
     future::Future,
     pin::Pin,
     str::FromStr,
@@ -11,7 +11,7 @@ use thiserror::Error;
 use tokio::sync::RwLock;
 use tracing::{debug, error, info, warn};
 use tycho_client::feed::{synchronizer::ComponentWithState, FeedMessage, Header};
-use tycho_core::Bytes;
+use tycho_core::{dto::ProtocolStateDelta, Bytes};
 
 use crate::{
     evm::{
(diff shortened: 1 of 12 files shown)


src/evm/protocol/vm/tycho_decoder.rs
```diff
@@ -115,7 +115,20 @@ impl TryFromWithBlock<ComponentWithState> for EVMPoolState<PreCachedDB> {
...
+            HashMap::new()
+        } else {
+            // enable triggering updates to this pool for all involved contracts
+            snapshot
+                .component
+                .contract_ids
+                .iter()
+                .map(|address| (address.clone(), id.clone()))
+                .collect()
+        };
```
zizou0x, line 131: I am not convinced that a breaking change on the interface (it's also weird to have something else than `Self` out of a `try_from`) is needed for this to be honest. Plus this also affects native implementation that are now returning this empty hashmap that isn't used anywhere.


Why does this logic need be done inside `try_from_with_block`? Can't it be directly implemented on `ComponentWithState`?

src/evm/decoder.rs
```diff
@@ -333,6 +349,65 @@ impl TychoStreamDecoder {
...
+                for pool in pools_to_update {
+                    match updated_states.entry(pool.clone()) {
+                        Entry::Occupied(mut entry) => {
+                            // if state exists in updated_states, update it
+                            let state: &mut Box<dyn ProtocolSim> = entry.get_mut();
+                            state
+                                .delta_transition(
+                                    ProtocolStateDelta::default(),
+                                    &state_guard.tokens,
+                                )
```
zizou0x, line 375: I find this weird, are we calling `delta_transition` with empty delta to trigger the update?

src/evm/protocol/vm/tycho_decoder.rs
louise-poole: It didn't make sense to me to be decoding parts of the ComponentWithState twice. Like we'll decode the manula updates static attribute and the contract ids part both inside and outside the vm decoder then. 
Also, the contracts to pool ids map is a vm specific thing so made sense to be in populated the vm decoder no?

src/evm/decoder.rs
louise-poole: Yeah, unfortunately with the nature of us using the same interfaces for VM and native implementations, theres info passed in both delta_transition and try_from_with_block that is ignored in certain cases.

src/evm/protocol/vm/tycho_decoder.rs
louise-poole: The higher level decoder is meant to be as generic as possible with these protocol-specific 'tycho-decoders' handling the actual value decoding and protocol specific stuff.

zizou0x: What is protocol specific here? I think this being in the decoder is like having the logic to update vm state there, it's only used by VMs but still pretty generic

louise-poole: Decoding for the 'manual_updates' and contract ids part is protocol specific (applies to certain VMs). 
Buuuut having another look here, the ProtocolComponent struct was recently expanded to include more than just the component id and tokens. We could generate this hashmap in the `Construct component from snapshot` section instead of in the 'Construct state from snapshot' section!

zizou0x (approved):
Awesome, thanks!

tvinagre (approved):

PR #149, opened 2025-02-19
feat: update uniswap v4 decoding of lp fees

louise-poole:
LP fees were moved from state attributes to static attributes.


zizou0x (approved):
LGTM

PR #155, opened 2025-02-27
fix(quickstart): Usability Fixes

Haikane:
- Interactive selection menus
- Refactored transaction code into reusable function
- Improved price and token amount formatting with human-readable values
- Added better display of swap prices showing rates in both directions to let user check if prices look good
- Error recovery for stream disconnections and simulation failures


dianacarvalho1 (approved):
thank yooouu

PR #160, opened 2025-03-04
feat(price_printer): display the protocol name for each row in the table

zizou0x:
This PR adds the protocol name for each protocol component in the price printer example.

It also fixes a small bug in the quickstart example.


louise-poole (approved):
Insane how little code this needed 😂 Nice!

PR #164, opened 2025-03-07
refactor: make protocol state fields public to access state data directly via ProtocolStreamBuilder

0xMerso:
Access to the internal field of protocol state structures is currently impossible when they are streamed by ProtocolStreamBuilder (although TychoStreamBuilder allows it).

The ProtocolStreamBuilder streams a BlockUpdate structure containing protocol states, each of which can be converted into a specific protocol state (UniswapV3State, etc.).

This PR changes the state structure fields from internal to public, which means that the attributes (liquidity ticks, etc.) can be read. Subject also mentioned in Tycho Telegram.

### DIFF
--- a/src/evm/protocol/uniswap_v3/state.rs
+++ b/src/evm/protocol/uniswap_v3/state.rs
@@ -32,11 +32,11 @@ use crate::{
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct UniswapV3State {
-    liquidity: u128,
-    sqrt_price: U256,
-    fee: FeeAmount,
-    tick: i32,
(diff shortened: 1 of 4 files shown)


0xMerso:
Seems that not all conflicts on the Github UI are displayed, so there is still a conflict. Sorry
I can just delete this PR, and make another fork and PR, maybe that would work better. Let me know.

louise-poole:
> Seems that not all conflicts on the Github UI are displayed, so there is still a conflict. Sorry I can just delete this PR, and make another fork and PR, maybe that would work better. Let me know.

Sure! If you're happy to do that then go for it 👍

PR #167, opened 2025-03-11
feat: add `get_limits` in `ProtocolSim`.

zizou0x:
This has been requested by clients that need to know the maximum amount out they can get out of the protocol component for their solving algorithm.


src/evm/protocol/vm/state.rs
```diff
@@ -259,16 +259,16 @@ where
         &self,
         tokens: Vec<Address>,
         overwrites: Option<HashMap<Address, HashMap<U256, U256>>>,
-    ) -> Result<U256, SimulationError> {
+    ) -> Result<(U256, U256), SimulationError> {
```
louise-poole, line 262: Why did you change this? This change doesn't appear to be used?

louise-poole: We can likely get rid of this fn now and just use `get_limits` instead?

src/evm/protocol/uniswap_v2/state.rs
```diff
@@ -100,6 +101,21 @@ impl ProtocolSim for UniswapV2State {
...
+        token_out: Address,
+    ) -> Result<(Option<BigUint>, Option<BigUint>), SimulationError> {
+        if self.reserve0 == U256::from(0u64) || self.reserve1 == U256::from(0u64) {
+            return Ok((Some(BigUint::zero()), Some(BigUint::zero())));
+        }
+
+        let zero_for_one = token_in < token_out;
+        let reserve_out = if zero_for_one { self.reserve1 } else { self.reserve0 };
+
+        Ok((None, Some(u256_to_biguint(reserve_out))))
```
louise-poole, line 116: This is introducing different behaviour from vm limits. On VM adapters the limit can be either soft or hard... soft limits are where the it's not advised to swap beyond this limit, however the simulation won't revert if you do. Hard limits are if the simulation will revert. Do we want to redefine that here? Or should we stick to the existing set up and return a sell limit with a soft limit?

zizou0x: Address in https://github.com/propeller-heads/tycho-simulation/pull/167/commits/153d4929173d182ce2265eb17307269479f14b2a after our discussion on slack

src/evm/protocol/vm/state.rs
zizou0x: We can't use `get_limits` instead of this one because of the interface difference. But I can reuse this one inside get_limits to avoid the code duplication

zizou0x: Addressed in https://github.com/propeller-heads/tycho-simulation/pull/167/commits/5675e1972c58eda5460a10ae1550244babbcb96a

zizou0x: We could get rid of this if we put the overwrite directly in `get_limits` but it's less flexible and we need to convert from `BigInt` into `U256` so I decided to reuse this one in `get_limits` instead

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -280,6 +281,92 @@ impl ProtocolSim for UniswapV4State {
...
+                    liquidity_math::add_liquidity_delta(current_liquidity, liquidity_delta);
+            }
+
+            if zero_for_one && current_tick < next_tick {
+                break;
+            }
+
+            if !zero_for_one && current_tick == next_tick {
+                break;
+            }
```
zizou0x, line 361: I don't like this, I think `next_initialized_tick_within_one_word` should return an error when we're below the smallest tick instead of this:

```
            if self.is_below_smallest(tick) {
                let minimum = cmp::max(self.ticks[0].index - spacing, min_in_word);
                return Ok((minimum, false));
            }
```

src/evm/protocol/uniswap_v2/state.rs
```diff
@@ -100,6 +101,30 @@ impl ProtocolSim for UniswapV2State {
...
+
+        let zero_for_one = token_in < token_out;
+        let (reserve_in, reserve_out) = if zero_for_one {
+            (self.reserve0, self.reserve1)
+        } else {
+            (self.reserve1, self.reserve0)
+        };
+
+        // Soft limit for amount in is the amount to get a 90% price impact.
+        // It is given by this: (√10 - 1) × reserve0 = 2.16 × reserve0
```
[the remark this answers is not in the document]
zizou0x: Addressed in https://github.com/propeller-heads/tycho-simulation/pull/167/commits/73d7a1ac86dbc6b51e23ffb66bff0bd59ed9b406

louise-poole (approved):

src/evm/protocol/utils/uniswap/sqrt_price_math.rs
```diff
@@ -31,7 +31,7 @@ fn div_rounding_up(a: U256, b: U256) -> Result<U256, SimulationError> {
     }
 }
 
-pub(super) fn get_amount0_delta(
+pub(crate) fn get_amount0_delta(
```
kayibal, line 34: Is this shared with v4 or why was this change necessary?

src/protocol/state.rs
```diff
@@ -104,6 +105,34 @@ pub trait ProtocolSim: std::fmt::Debug + Send + Sync + 'static {
...
+    /// This function calculates the maximum possible trade amount between two tokens,
+    /// taking into account the protocol's specific constraints and mechanics.
+    /// The implementation details vary by protocol - for example:
+    /// - For constant product AMMs (like Uniswap V2), this is based on available reserves
+    /// - For concentrated liquidity AMMs (like Uniswap V3), this considers liquidity across tick
+    ///   ranges
+    ///
+    /// Note: if there are no limits, the returned amount will be a "soft" limit,
+    ///       meaning that the actual amount traded could be higher but it's advised to not
+    ///       exceed it.
```
kayibal, line 119: We should commit to the 90% or make them configurable and then mention this here. If not the user doesn't know what to expect here...

src/protocol/state.rs
```diff
@@ -104,6 +105,34 @@ pub trait ProtocolSim: std::fmt::Debug + Send + Sync + 'static {
...
+    ///
+    /// # Returns
+    /// * `Ok((Option<BigUint>, Option<BigUint>))` - A tuple containing:
+    ///   - First element: The maximum input amount
+    ///   - Second element: The maximum output amount
+    /// * `Err(SimulationError)` - If any unexpected error occurs
+    fn get_limits(
+        &self,
+        sell_token: Address,
+        buy_token: Address,
```
kayibal, line 133: Would probably make this a struct so we can later add the 90% as a parameter here for soft limits

src/protocol/state.rs
```diff
@@ -104,6 +105,34 @@ pub trait ProtocolSim: std::fmt::Debug + Send + Sync + 'static {
...
+    ///       exceed it.
+    ///
+    /// # Arguments
+    /// * `sell_token` - The address of the token being sold
+    /// * `buy_token` - The address of the token being bought
+    ///
+    /// # Returns
+    /// * `Ok((Option<BigUint>, Option<BigUint>))` - A tuple containing:
+    ///   - First element: The maximum input amount
+    ///   - Second element: The maximum output amount
```
kayibal, line 128: I would explain this via domain aka input range for two different functions. `[0, res.0]` is the input domain for `get_amount_out`outside this range the function is "not defined". `[0, res.1]` it is the input domain for `get_amount_in` (we don't expose this yet on this interface).

src/evm/protocol/uniswap_v2/state.rs
```diff
@@ -100,6 +101,34 @@ impl ProtocolSim for UniswapV2State {
...
+        // Soft limit for amount in is the amount to get a 90% price impact.
+        // The two equations to resolves are:
+        // 90% price impact: (reserve1 - y)/(reserve0 + x) = 0.1 × (reserve1/reserve0)
+        // Maintain constant product: (reserve0 + x) × (reserve1 - y) = reserve0 * reserve1
+        //
+        // This resolves into x = (√10 - 1) × reserve0 = 2.16 × reserve0
+        let amount_in =
+            safe_div_u256(safe_mul_u256(reserve_in, U256::from(216))?, U256::from(100))?;
+
+        Ok((u256_to_biguint(amount_in), u256_to_biguint(reserve_out)))
```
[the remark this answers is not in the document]
zizou0x: It depends how we want to design soft limits, it's still not 100% clear to me so I'll follow your intuition and make it 90% price impact limit on the amount out too

src/protocol/state.rs
zizou0x: I'm not sure what to do here, in our slack discussion we decided that https://github.com/propeller-heads/tycho-simulation/pull/167/commits/4a607cbd6b9a5e1c9b9db48881015a50de671779 wasn't good because it's not consistent with the vm implementation. If we now add the price impact in the interface then we end up having the same issue with inconsistencies

src/protocol/state.rs
zizou0x: I am hesitant to do that for the reason described [above](https://github.com/propeller-heads/tycho-simulation/pull/167#discussion_r2005321414)

PR #171, opened 2025-03-19
feat: extend uniswap v3 fees to support pancakeswap v3

louise-poole:
Pancakeswap V3 simulation logic is identical to that of Uniswap V3, however it supports more fee tiers. 
Pancakeswap V3 can now be streamed and simulated on using the `UniswapV3State` implementation.


zizou0x (approved):
Looks perfect!

To be picky: I would have preferred this test to have been done with a pool with a new fee tier. But I think it's good enough like this, I don't see how a different tier would break the logic

tvinagre (commented):
I'd prefer if this was done in a separate Pool Type, if possible - since Pancake is not a 1:1 clone to UniV3, it can confuse users. I'm fine with adding simply a note, either in the code or on the docs to point the users to use UniV3 struct when dealing with PancakeV3, but let's add an issue to separate them, please.

PR #172, opened 2025-03-19
feat: update tycho-execution version

tamaralipows:
- This update is needed to fix a slippage issue where 0.1% would be rounded to 0
- The latest tycho-execution also supports cyclical swaps, although this new feature isn't needed in the latest quickstart


royvardhan (approved):
thanks!

PR #173, opened 2025-03-20
feat: Ekubo Integration

die-herdplatte:
In conjunction with https://github.com/propeller-heads/tycho-protocol-sdk/pull/172 and https://github.com/propeller-heads/tycho-execution/pull/126


src/evm/protocol/ekubo/state.rs
```diff
@@ -0,0 +1,251 @@
...
+#[enum_delegate::implement(EkuboPool)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EkuboState {
+    Base(BasePool),
+    Oracle(OraclePool),
+}
+
+impl EkuboState {
+    // Minimum overhead of initiating any swap with the Core contract
+    const BASE_GAS_COST: u64 = 50_000;
```
die-herdplatte, line 36: Should this include gas costs of token transfers?

zizou0x: No, gas cost of token transfers are accounted for separately

src/evm/protocol/ekubo/tycho_decoder.rs
```diff
@@ -0,0 +1,249 @@
...
+#[derive(Error, Debug)]
+pub enum StateDecodingError {
+    #[error(transparent)]
+    InvalidSnapshot(#[from] InvalidSnapshotError),
+    #[error("unsupported extension")]
+    UnsupportedExtension,
+}
+
+impl TryFrom<Bytes> for EkuboExtension {
+    type Error = StateDecodingError;
```
zizou0x, line 40: It would be nice to make this a `InvalidSnapshotError`, it's currently a requirement for the builder see [here](https://github.com/propeller-heads/tycho-simulation/blob/e588151d25a6d8070f7813b4ea8608329d221ab2/src/evm/stream.rs#L74)

src/evm/protocol/ekubo/state.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+fn sqrt_price_q128_to_f64(x: U256, (token0_decimals, token1_decimals): (usize, usize)) -> f64 {
+    let token_correction = 10f64.powi(token0_decimals as i32 - token1_decimals as i32);
+
+    let price = u256_to_f64(alloy_primitives::U256::from_limbs(x.0)) / 2.0f64.powi(128);
+    price.powi(2) * token_correction
+}
+
+impl ProtocolSim for EkuboState {
```
zizou0x, line 69: We added a new function called `get_limits` here. It's used by some solvers to optimize their logic, can you add it please?

zizou0x (commented):
Looks very good, I tested the integration with the indexer and it worked nicely. 👍 

Just two small comments before we can merge this.

zizou0x (approved):
Looks nice! I added some small questions that can be addressed in a later PR. Nothing concerning so I'll merge this one

src/evm/protocol/ekubo/pool/base.rs
```diff
@@ -0,0 +1,193 @@
...
+        self.state.liquidity = liquidity;
+    }
+
+    fn reinstantiate(&mut self) {
+        // Only after a swap we set the active_tick to None. In this case, the active_tick_index is
+        // already correctly computed though
+        if let Some(active_tick) = self.active_tick {
+            self.state.active_tick_index =
+                find_nearest_initialized_tick_index(self.ticks.inner(), active_tick);
+        }
```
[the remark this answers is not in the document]
die-herdplatte: The `active_tick_index` may change when we cross an initialized tick, therefore it's not possible to initialize it at creation and not change it ever again.
It would be possible to compute the `active_tick_index` directly in `set_active_tick` though and avoid the state variable.
Should I do that?

Also, as it is right now, the active tick may not have changed and we, nevertheless, recompute the tick index.
This could be optimized if you don't want to go for the former approach.

PR #179, opened 2025-03-27
fix: get tycho client & common deps from crates.io

kayibal:
This avoids dependency resultion issues in the future where we have more than one version of a crate pulled.

### DIFF
--- a/src/evm/tycho_models.rs
+++ b/src/evm/tycho_models.rs
@@ -3,7 +3,7 @@ use std::{collections::HashMap, fmt::Display};
 use alloy_primitives::{Address, B256, U256};
 use chrono::{NaiveDateTime, Utc};
 use serde::{Deserialize, Serialize};
-pub use tycho_core::{dto::ChangeType, models::Chain};
+pub use tycho_common::{dto::ChangeType, models::Chain};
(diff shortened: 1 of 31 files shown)


src/evm/tycho_models.rs
```diff
@@ -135,8 +135,8 @@ impl AccountUpdate {
     }
 }
 
-impl From<tycho_core::dto::AccountUpdate> for AccountUpdate {
-    fn from(value: tycho_core::dto::AccountUpdate) -> Self {
+impl From<tycho_common::dto::AccountUpdate> for AccountUpdate {
```
zizou0x, line 138: Maybe not in this PR scope but would be nice to import directly instead of using the full path

src/lib.rs
```diff
@@ -10,7 +10,8 @@ extern crate core;
 
 // Reexports
 pub use tycho_client;
-pub use tycho_core;
+pub use tycho_common;
+pub use tycho_common as tycho_core; // Use `tycho_common` directly instead of `tycho_core`.
```
zizou0x, line 14: This is a todo?

zizou0x (approved):

PR #185, opened 2025-04-01
refactor: update error handling in Token conversion

zizou0x:
This removes the panic in the TryFrom in the case where gas is missing


src/models.rs
```diff
@@ -82,13 +82,15 @@ impl Hash for Token {
 }
 
 impl TryFrom<ResponseToken> for Token {
-    type Error = std::num::TryFromIntError;
+    type Error = InvalidSnapshotError; //TODO: add a specific error
```
louise-poole, line 85: Why not make the specific error now quick?

zizou0x: Because I didn't want to get sidetracked from my main task, but I'm gonna do it now because I don't wanna leave todos behind me

louise-poole (approved):
Absolute perfection 👌

PR #187, opened 2025-04-01
feat: update tycho-execution version

tamaralipows:
- We no longer pass router address in the Solution object


dianacarvalho1 (approved):
Thank you @tamaralipows ! 🙏🏼 Don't forget to update the execution version to the last one though

PR #191, opened 2025-04-03
refactor: improvemenets to quickstart

tamaralipows:
- These improvements were copied over from the price printer. After this, we will be able to swap on more protocols if ethereum chain is chosen (sushiv2, pancakev2, pancakev3, balancer, ekubo). Before, we were restricting ourselves to protocols that were available on all chains.
- It also fetches the correct tycho default URL depending on chain.

### DIFF
--- a/examples/quickstart/main.rs
+++ b/examples/quickstart/main.rs
@@ -29,6 +29,7 @@ use num_bigint::BigUint;
 use num_traits::ToPrimitive;
 use tracing_subscriber::EnvFilter;
 use tycho_common::Bytes;
+pub mod utils;
 use tycho_execution::encoding::{
     evm::{
         encoder_builder::EVMEncoderBuilder, tycho_encoder::EVMTychoEncoder, utils::encode_input,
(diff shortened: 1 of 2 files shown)


examples/quickstart/main.rs
```diff
@@ -114,14 +125,52 @@ async fn main() {
...
-        .exchange::<UniswapV3State>("uniswap_v3", tvl_filter.clone(), None)
-        .exchange::<UniswapV4State>(
-            "uniswap_v4",
-            tvl_filter.clone(),
-            Some(uniswap_v4_pool_with_hook_filter),
-        )
+    let mut protocol_stream = ProtocolStreamBuilder::new(&tycho_url, chain);
+
+    match chain {
+        Chain::Ethereum => {
```
dianacarvalho1, line 131: maybe we could make this general too and use it in both examples?? or do you think it's an overkill?

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 only have 1 question!

examples/quickstart/main.rs
tamaralipows: you mean just this match statement? I think it's not the best idea since execution may at times support fewer protocols than simulation

dianacarvalho1: I mean in the quickstart and the price printer!

dianacarvalho1 (approved):
✨ ✨ ✨

PR #194, opened 2025-04-09
fix(simulation): fix balances overwrites bugs

zizou0x:
This PR fixes two bugs in the balance overwrites:

- We were using account-balances even when they were empty
- We were deleting some useful balances on updates for >2 tokens pools

Also added EkuboV2 in the price printer and updated the readme for Unichain support


dianacarvalho1 (approved):
Thank yooouuu @zizou0x 🙏🏼 🙏🏼

PR #196, opened 2025-04-12
feat: Ekubo TWAMM & MEV-resist integration

die-herdplatte:
Integrates Ekubo [TWAMM](https://docs.ekubo.org/user-guides/dollar-cost-average-orders) and MEV-resist pools


src/evm/protocol/ekubo/pool/twamm.rs
```diff
@@ -0,0 +1,291 @@
...
+        self.changed_virtual_order_deltas
+            .push(delta);
+    }
+
+    fn estimate_block_timestamp(&self) -> u64 {
+        if self.swapped_this_block {
+            self.state.last_execution_time
+        } else {
+            // TODO How accurate is it to take the current timestamp?
+            Ord::max(self.state.last_execution_time + SLOT_DURATION_SECS, current_timestamp())
```
die-herdplatte, line 112: Any ideas on good approximations here?

src/evm/protocol/ekubo/attributes.rs
```diff
@@ -0,0 +1,92 @@
...
+                    "orders attribute should have 3 segments but received {splits_len}"
+                )));
+            }
+
+            let time: u64 = match splits[2].parse() {
+                Ok(time) => time,
+                Err(err) => return Some(Err(format!("order time can't be parsed as u64: {err}"))),
+            };
+
+            if time <= last_execution_time {
```
die-herdplatte, line 59: Can you think of any issues with reorgs here?

src/evm/protocol/ekubo/pool/twamm.rs
```diff
@@ -0,0 +1,291 @@
...
+            swapped_this_block: false,
+        })
+    }
+
+    pub fn last_execution_time(&self) -> u64 {
+        self.state.last_execution_time
+    }
+
+    pub fn set_last_execution_time(&mut self, last_execution_time: u64) {
+        self.state.last_execution_time = last_execution_time;
```
die-herdplatte, line 91: Potential to remove old virtual order deltas here.
But again, unsure about implications of reorgs.

src/evm/protocol/ekubo/pool/twamm.rs
zizou0x: It's unclear to me what this should be? Does it ideally need to be the timestamp when the swap is executed?

How does this affect the swap? It can affect it in both direction i.e positively and negatively?

src/evm/protocol/ekubo/attributes.rs
zizou0x: Depending on how `last_execution_time` is retrieved yes we could have issues.

From what I see we always retrieve it from the attributes before calling this. Tycho handles reorgs atomically, meaning if we always use the latest `last_execution_time` we should be good

src/evm/protocol/ekubo/pool/twamm.rs
zizou0x: This has been removed right?

zizou0x (commented):
Thanks!

I tested this on my machine and it seems to work. Just added a few questions to make sure I'm not missing some edge cases

src/evm/protocol/ekubo/attributes.rs
die-herdplatte: Sounds good

src/evm/protocol/ekubo/pool/twamm.rs
kayibal: Ideally this pool needs to have access to the latest blocks timestamp.... unfortunately we don't have a way to access this easily yet. And then we should calculate `latest_block.ts + SLOT_DURATION_SECS`

src/evm/protocol/ekubo/pool/twamm.rs
die-herdplatte: Done

zizou0x (approved):
LGTM! Thanks a lot for this one!

PR #203, opened 2025-04-21
feat: move CPMM core logic to dedicated CPMM protocol trait

louise-poole:
This allows us to support CPMM protocols with different fees without duplicating core logic. This is currently divided into 2 implementations of the trait:
- UniswapV2 (0.3% fee)
- PancakeswapV2 (0.25% fee)


zizou0x (approved):
Overall looks good, just please add back the comments if they are still relevant

src/evm/protocol/cpmm/protocol.rs
```diff
@@ -25,17 +25,11 @@ pub trait CPMMProtocol {
...
 
-    /// Get the reserve of token 1
-    fn get_reserve1(&self) -> U256;
-
-    /// Get mutable reference to reserve of token 0
-    fn get_reserve0_mut(&mut self) -> &mut U256;
-
-    /// Get mutable reference to reserve of token 1
-    fn get_reserve1_mut(&mut self) -> &mut U256;
+    /// Get mutable reference to reserves of both tokens
```
zizou0x, line 31: Can we specify in the doc that it should be (reserve0,reserve1) and not the opposite please?

PR #221, opened 2025-05-28
feat: Initial draft for hooks simulation

dianacarvalho1:


src/evm/protocol/uniswap_v4/utils.rs
```diff
@@ -0,0 +1,27 @@
...
+    BeforeDonate = 5,
+    AfterSwap = 6,
+    BeforeSwap = 7,
+    AfterRemoveLiquidity = 8,
+    BeforeRemoveLiquidity = 9,
+    AfterAddLiquidity = 10,
+    BeforeAddLiquidity = 11,
+    AfterInitialize = 12,
+    BeforeInitialize = 13,
+}
```
tamaralipows, line 20: Are these the only options? How are these extracted/categorized? What if a hook wants to do multiple of these, or something not included?

dianacarvalho1: yes these are the only options. I'm not sure how but this is all added when making the hook address. but we are only interested in decoding this (I took this from [here](https://github.com/shuhuiluo/uniswap-v4-sdk-rs/blob/main/src/utils/hook.rs#L69))

tamaralipows: ohhh so basically we reject all hooks that dont fall into these

src/evm/simulation.rs
```diff
@@ -114,24 +117,31 @@ where
             ..Default::default()
         };
 
-        let mut cfg_env: CfgEnv<SpecId> = CfgEnv::default();
+        let mut cfg_env: CfgEnv<SpecId> = CfgEnv::new_with_spec(SpecId::PRAGUE);
```
tamaralipows, line 120: What does this do

dianacarvalho1: it changes the version of the EVM we are using (related to the upgrades) 
but this change here does literally nothing 😅 sorry. because the default one is Prague

PR #228, opened 2025-06-06
feat: Add skeleton of GenericVMHookHandler

dianacarvalho1:
Move hook_handler.rs to inside hooks directory
Add pool_manager_bytecode.bin and use it in GenericVMHookHandler


src/evm/protocol/uniswap_v4/hooks/pool_manager_bytecode.bin
tamaralipows, line 1: How on earth did you save to this file

PR #231, opened 2025-06-09
feat: Implement after_swap in generic_vm_hook_handler

dianacarvalho1:
extra: Move bytecode bins into an assets directory


src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -269,4 +309,64 @@ mod tests {
...
+            swap_params,
+            delta: I256::from_dec_str("-3777134272822416944443458142492627143113384069767150805")
+                .unwrap(),
+            hook_data: Bytes::new(),
+        };
+
+        let result = hook_handler.after_swap(after_swap_params, block.number);
+
+        let res = result.unwrap().result;
+        assert_eq!(res, I256::from_raw(U256::from_str("0").unwrap()));
```
tamaralipows, line 370: Which transaction is this simulating? Can you include this here? Otherwise how do we know this value is right?

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -152,11 +151,52 @@ where
...
+            None,
+            Some(self.pool_manager),
+            U256::from(0u64),
+            Some(transient_storage_params),
+        )?;
+
+        let decoded = AfterSwapReturn::abi_decode(&res.return_value).map_err(|e| {
+            SimulationError::FatalError(format!("Failed to decode before swap return value: {e:?}"))
+        })?;
+        Ok(WithGasEstimate {
```
tamaralipows, line 194: Should we add a TODO to verify the problem that we encountered with transient storage in other contracts? Or open another task for it?

PR #232, opened 2025-06-09
feat(GenericVMHookHandler): Return errors for unimplemented methods

tamaralipows:
- these errors are considered recoverable in this case, since Tycho simulation is the one that will be catching the error if the hook is not implemented in the UniswapV4PoolState. This is a completely valid state and will just signal for Tycho simulation to not call these methods for the hook.


dianacarvalho1 (approved):
peerfeeectt 👌🏼

PR #234, opened 2025-06-09
feat: HookHandlerCreator methods

tamaralipows:
This PR combines three tasks:

- [x] [Implement `register_hook_handler`](https://propeller-heads.atlassian.net/browse/ENG-4566)
- [x] [Implement `set_default_hook_handler`](https://propeller-heads.atlassian.net/browse/ENG-4567)
- [x] [Implement `instantiate_hook_handler`](https://propeller-heads.atlassian.net/browse/ENG-4568)

These were so small it was easier to put them in one PR instead of going back and forth between several PRs to see how they played together. I also needed to touch all of them so I could get a better understanding here.

**TODO:**
**Where should the handlers actually be registered?**

The  thing I've considered is to do this when we are registering exchanges, though this doesn't seem right, since I need the hook address, which I don't think I have until the decoding point...

**EDIT:** The hook address will be hardcoded when we register exchanges.


src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
```diff
@@ -44,8 +44,45 @@ pub trait HookHandlerCreator: Send + Sync {
...
+    handler: Box<dyn HookHandlerCreator>,
+) -> Result<(), SimulationError> {
+    HANDLER_FACTORY
+        .write()
+        .map_err(|e| SimulationError::FatalError(e.to_string()))?
+        .insert(hook, handler);
+    Ok(())
+}
+
+pub fn set_default_hook_handler(
```
tamaralipows, line 81: I am having trouble understanding when we would ever need to set this. I am assuming we don't, since it's already set to `GenericVMHookHandlerCreator` at start time?

src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
```diff
@@ -44,8 +44,62 @@ pub trait HookHandlerCreator: Send + Sync {
     ) -> Result<Box<dyn HookHandler>, InvalidSnapshotError>;
 }
 
+pub struct GenericVMHookHandlerCreator;
```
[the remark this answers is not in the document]
dianacarvalho1: what VM dir? 🤔

src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
```diff
@@ -44,8 +44,62 @@ pub trait HookHandlerCreator: Send + Sync {
...
+
 // Workaround for stateless decoder trait.
+// Mapping from hook address to the handler creator.
 lazy_static! {
     static ref HANDLER_FACTORY: RwLock<HashMap<Address, Box<dyn HookHandlerCreator>>> =
         RwLock::new(HashMap::new());
 }
+
+lazy_static! {
+    static ref DEFAULT_HANDLER: RwLock<Box<dyn HookHandlerCreator>> =
```
dianacarvalho1, line 66: why does it need a `RwLock`?

src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
dianacarvalho1: ah if you are doing the `DEFAULT_HANDLER`, then yes we don't need this!

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 
I think this PR is complete as is tbh 🤔 
If we had other hook implementations (that not the default/VM one) we would need to indeed call the register function. 
It can't be when we decode though. And where are we registering exchanges? But we do know the hook address. This needs to be hardcoded by the developers.

src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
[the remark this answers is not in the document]
tamaralipows: Would it not make sense to put this in the vm dir as opposed to the uniswapv4 dir? Like the same dir where we put the generic vm hook handler?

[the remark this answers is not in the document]
dianacarvalho1: didn't we agree to put everything inside the uniswapv4 dir [here](https://propellerswap.slack.com/archives/C05LQALE9L1/p1749221948804739)? 🤔

src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
tamaralipows: Will it never be that multiple threads attempt to set the default creator at once? I assumed if that were the case for the factory, that it could also be the case here?

PR #235, opened 2025-06-10
feat: GenericVMHookHandlerCreator

tamaralipows:


src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
```diff
@@ -49,9 +68,82 @@ pub struct GenericVMHookHandlerCreator;
...
+            timestamp: params.block.timestamp,
+        };
+
+        let db = SimulationDB::new(get_client(None), get_runtime(), Some(block_header));
+
+        let engine = create_engine(db, true).map_err(|e| {
+            InvalidSnapshotError::VMError(SimulationError::FatalError(format!(
+                "Failed to create engine: {e:?}"
+            )))
+        })?;
```
dianacarvalho1, line 101: no no, we don't need this. We should use the `SHARED_TYCHO_DB`

src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
```diff
@@ -49,9 +68,82 @@ pub struct GenericVMHookHandlerCreator;
...
+                token_address,
+                AccountInfo {
+                    balance: U256::ZERO, // Token contracts have zero ETH balance
+                    nonce: 1,
+                    code_hash,
+                    code: Some(erc20_bytecode),
+                },
+                None,
+                false,
+            );
```
dianacarvalho1, line 140: why are you doing this?? 🤔

dianacarvalho1: I'm not so sure that this is needed and we talked about this the other day remember? 🤔 and we said: let's not do it until we are sure we need it.
and if this is needed I think this should be done inside the GenericVMHookHandler actually, not here (so we would have to change the GenericVMHookHandler interface)

dianacarvalho1: One thing that is for sure missing is initing the stateless_contracts (it should be at least the pool manager)

dianacarvalho1: huummm I see that for the EVMPoolState we are actually doing this in the builder though. meh I won't insist on where we should do this

dianacarvalho1: I'm not sure about the balances thingy (or even what they mean) but for sure we need the stateless contracts and maybe the tokens (I think it will depend on the hook)

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 I left some comments!

src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
tamaralipows: @dianacarvalho1 I'm a bit confused by this message... what is the point of the params if none of these attributes are needed after all? We aren't passing the params to the handler so they can't really be used there... having a hard time understanding this class 

I don't remember this conversation - without having actually started this task it's hard for me to remember relevant convos if we didn't write them down :/

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -63,10 +63,15 @@ where
 {
     pub fn new(
         address: Address,
-        bytecode: Bytecode,
         engine: SimulationEngine<D>,
         pool_manager: Address,
+        // TODO are all of these necessary?
+        _all_tokens: HashMap<Bytes, Token>,
+        _account_balances: HashMap<Bytes, HashMap<Bytes, Bytes>>,
```
dianacarvalho1, line 70: I would say that account_balances is enough. it should include already this component's balance _I think_

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -63,10 +63,15 @@ where
...
+        _balances: HashMap<Bytes, Bytes>,
     ) -> Result<Self, SimulationError> {
+        // TODO overwrite token balances (see how it's done in USV4 Pool State)
```
dianacarvalho1, line 73: it's not done in the USV4 Pool State. it's done in the EVM pool state

src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
```diff
@@ -18,9 +24,12 @@ pub struct HookCreationParams<'a> {
     all_tokens: &'a HashMap<Bytes, Token>,
     state: UniswapV4State,
     /// Attributes of the component. If an attribute's value is a `bigint`,
-    /// it will be encoded as a big endian signed hex string.
+    /// it will be encoded as a big endian signed hex string. See ResponseProtocolState for more
+    /// details.
     pub(crate) attributes: &'a HashMap<String, Bytes>,
-    /// Sum aggregated balances of the component.
+    // TODO are these balances a mapping from storage slot to value? I can't find an example in
```
dianacarvalho1, line 30: can you ask cairoline please?

dianacarvalhoo (approved):
Thank you 🙏

PR #236, opened 2025-06-11
fix(python): Do not error on account balances for non-pool tokens

louise-poole:
Account balances include balances for all tokens held by that account, regardless of if that token is associated with the current pool being decoded. Since the python decoder reports balances in off-chain amounts, it's important that we have the token object for all balances we're processing so that we can do this conversion. Therefore, to avoid issues we should only process account balances for tokens that belong to the pool being decoded.


zizou0x (approved):
LGTM as a temporary fix

Looking forward to have Rust bidings to remove this

PR #238, opened 2025-06-11
feat: Implement has_permission for hooks

tamaralipows:
[Issue](https://propeller-heads.atlassian.net/browse/ENG-4573)

Tests taken from here: https://github.com/shuhuiluo/uniswap-v4-sdk-rs/blob/main/src/utils/hook.rs#L93


dianacarvalho1 (approved):
uuuhh very good! thank you 🙏🏼

PR #244, opened 2025-06-24
feat: improve price-printer ux

zizou0x:
This PR improve the UX for the price-printer example. Specifically:

- Added a new functionality allowing users to input the swap amount
- Added a new line in the quote popup to display the swap direction


louise-poole (approved):
🙏

PR #245, opened 2025-06-25
feat: RFQ draft

greendianasaur:
Created its own tycho-rfq package for this.

Introduced a new trait `IndicativelyPriced` (on the pool state level). `IndicativelyPriced` trait marks protocols that support indicative pricing and optionally binding quotes. This way the user can call `request_binding_quote` at any time after it gets the `RFQState`s

Each RFQ has:  
1. A client that connects to their API and has two methods: one to fetch indicative prices (levels, formulas, etc) and another to get a binding quote
2. A State that implements ProtocolSim and IndicativelyPriced
3. A decoder that reads `ComponentWithState` and creates the Pool state

At a higher level we have a `RFQStreamBuilder`:
1. that uses the `RFQClient`s streams, that output `StateSyncMessage`s.
2. it combines all the streams and constructs a `FeedMessage`
3. uses a general `RFQDecoder` that converts the `FeedMessage` into `Update` by using the individual RFQ decoders

TODO: how to handle errors or stale streams?

### Execution notes (TODO)
Most RFQs allow to set the sender and receiver to whatever we want (this must be done when requesting the signed quote). We only know the sender and the receiver at encoding time, so ideally the request for quote needs to be done inside encoding. This means mega refactor. Encoding needs to be able to access the ProtocolSim implementations. We can do that by moving some structs and traits from tycho-simulation to tycho-common.

image

### DIFF
--- a/tycho_rfq/src/stream.rs
+++ b/tycho_rfq/src/stream.rs
@@ -3,65 +3,40 @@ use std::{collections::HashMap, pin::Pin, sync::Arc};
 use chrono::Utc;
 use futures::{StreamExt, stream::select_all};
 use tycho_simulation::{
-    protocol::{errors::InvalidSnapshotError, models::Update, state::ProtocolSim},
-    tycho_client::feed::{FeedMessage, synchronizer::StateSyncMessage},
+    evm::decoder::TychoStreamDecoder,
+    protocol::{
+        errors::InvalidSnapshotError,
+        models::{TryFromWithBlock, Update},
+        state::ProtocolSim,
+    },
+    tycho_client::feed::{
+        FeedMessage,
+        synchronizer::{ComponentWithState, StateSyncMessage},
+    },
 };
 
 use crate::client::RFQClient;
 
-type DecodeFut =
-    Pin<Box<dyn Future<Output = Result<Box<dyn ProtocolSim>, InvalidSnapshotError>> + Send + Sync>>;
-type RegistryFn = dyn Fn(StateSyncMessage) -> DecodeFut + Send + Sync;
-pub struct RFQDecoder {
-    registry: HashMap<String, Box<RegistryFn>>,
-}
-
-impl RFQDecoder {
-    pub async fn decode(&self, msg: FeedMessage) -> Update {
-        // Something like:
-        let mut states = HashMap::new();
-        for (provider, state_msg) in msg.state_msgs {
-            // Create ProtocolStates for all pairs
-            // we will never have a partial update, so we need to construct the states over and over
-            // again here
-            if let Some(decode_fn) = self.registry.get(&provider) {
-                match decode_fn(state_msg).await {
-                    Ok(sim) => {
-                        states.insert(provider.clone(), sim);
-                    }
-                    Err(err) => {
-                        eprintln!("Error decoding {}: {:?}", provider, err);
-                    }
-                }
-            } else {
-                eprintln!("No decoder registered for provider: {}", provider);
-            }
-            // gather all deleted pairs
-        }
-
-        let timestamp = Utc::now().timestamp() as u64;
-
-        Update {
-            block_number: timestamp,
-            states,
-            new_pairs: Default::default(),
-            removed_pairs: Default::default(),
-        }
-    }
-}
-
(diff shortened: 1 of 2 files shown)


tycho_rfq/src/protocols/bebop/client.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        // example snapshot for Bebop
+        let snapshot = Snapshot {
+            states: HashMap::from([(
+                id.clone(),
+                ComponentWithState {
+                    state: ResponseProtocolState {
+                        component_id: id.clone(),
+                        attributes: HashMap::new(), /* TODO: put the bids and asks here?? then we
+                                                     * would have to convert them to Bytes? */
```
greendianasaur, line 61: @kayibal I think this would be the only way to use the tycho client models.. what do you think?

greendianasaur: also.. these are dto models.. is it ok to use them?

src/protocol/models.rs
```diff
@@ -54,6 +54,7 @@ pub struct ProtocolComponent {
     pub static_attributes: HashMap<String, Bytes>,
     pub creation_tx: Bytes,
     pub created_at: NaiveDateTime,
+    pub additional_execution_data: Option<HashMap<String, Bytes>>,
```
kayibal, line 57: Can you explain a bit more here? I don't quite get what the key of the map here would be.

Also I see further down that there is mutable access to this attribute. I would prefer if ProtocolComponent data stays static.

src/protocol/models.rs
```diff
@@ -177,13 +189,19 @@ pub struct BlockUpdate {
...
+impl Update {
     pub fn new(
-        block_number: u64,
+        marker: u64,
         states: HashMap<String, Box<dyn ProtocolSim>>,
         new_pairs: HashMap<String, ProtocolComponent>,
     ) -> Self {
-        BlockUpdate { block_number, states, new_pairs, removed_pairs: HashMap::new() }
+        Update {
+            marker: marker.clone(),
```
kayibal, line 199: This clone seems unnecessary

src/protocol/models.rs
```diff
@@ -167,7 +178,8 @@ impl GetAmountOutResult {
 }
 
 #[derive(Debug, Clone)]
-pub struct BlockUpdate {
+pub struct Update {
+    pub marker: u64,
```
kayibal, line 182: Can you explain the reasoning here? Why did you do this change? What exactly is `marker` I assume it is a timestamp if we don't have a block number? If that is the case what would block number be set to?

tycho_rfq/src/protocols/bebop/client.rs
```diff
@@ -0,0 +1,105 @@
...
+impl RFQClient for BebopClient {
+    fn stream(&self) -> BoxStream<'static, (String, StateSyncMessage)> {
+        // connect to Bebop websocket
+        // for each message received
+        //   compare the token pairs that were received with the current ones.
+        //   define the removed ones (if any)
+        //
+
+        // example of a protocol component for a Bebop pair
+        let id = "WETH/USDC_maker1".to_string();
```
kayibal, line 39: It would be nice to be able to tvl filter these here

tycho_rfq/src/stream.rs
kayibal, line 1: This looks very similar to the existing StreamBuilder, why do we have to duplicate this here? Is it possible to make the existing StreamBuilder more flexible to accomodate this use case? Just asking might not be possible...

src/protocol/models.rs
greendianasaur: ai ups, this is from the previous version we did last cooldown. I think in the current plan we don't really need this

tycho_rfq/src/protocols/bebop/client.rs
greendianasaur: yes yes, I'm still missing that part and filtering for token pairs

greendianasaur: I think it would be best to filter these straight in the RFQClient!

PR #250, opened 2025-07-07
feat(univ4): update methods to use hook if available; create hook handler in tycho decoder

greendianasaur:


tamaralipows (approved):
Thank you very much!

PR #255, opened 2025-07-09
feat: Remove duplicated models

dianacarvalho1:
Done:
- Use Token from tycho-common
- Use ExtractorIdentity from tycho common instead of redefining it
- Use Block from tycho common instead of redefining it
- Remove unnecessary impl From<Block> for BlockHeader]
- Use Header from tycho-client instead of redefining a new BlockHeader
- Use ContractId from tycho-common instead of redefining
- Remove unused structs from tycho_models.rs

Corresponding PRs:
- https://github.com/propeller-heads/tycho-indexer/pull/618
- https://github.com/propeller-heads/tycho-execution/pull/239

See [task](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-4705) for a summary of how models connect in all three repos


src/evm/tycho_models.rs
```diff
@@ -1,115 +1,14 @@
...
-        extractor: String,
-        chain: Chain,
-        block: Block,
-        account_updates: HashMap<Address, AccountUpdate>,
-        new_pools: HashMap<Address, SwapPool>,
-    ) -> Self {
-        Self { extractor, chain, block, account_updates, new_pools }
-    }
-}
-
```
dianacarvalho1, line 112: now in this file we only have AccountUpdate  and ResponseAccount. I actually think they should be in tycho-common even though they are used only here 🤔 they are just a wrapper around the dto objects. what do you think?

PR #257, opened 2025-07-11
feat: Generalise Header and remove ProtocolSim

dianacarvalho1:
Done:
- Rename Header to BlockHeader from tycho-common
- Use ProtocolSim from tycho-common
- Refactor CPMMProtocol trait into util functions

Corresponding PRs:
- https://github.com/propeller-heads/tycho-indexer/pull/619
- https://github.com/propeller-heads/tycho-execution/pull/240

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8248,8 +8248,9 @@ dependencies = [
 
 [[package]]
 name = "tycho-client"
-version = "0.75.0"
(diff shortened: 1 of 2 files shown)


tamaralipows (commented):
Looking good so far! I still think option 3 you suggested is a good way to go!! Thank you!

tamaralipows (approved):
Thank you! I think the util function solution looks quite clean!

PR #259, opened 2025-07-14
feat: Add rfq feature gate

dianacarvalho1:
Create a feature gate rfq
Implement initial traits: IndicativelyPriced and RFQClient; and models: TimestampHeader, SignedQuote and RFQError


tamaralipows (approved):
Thank you! I think this is fine!

PR #261, opened 2025-07-15
feat: (WIP) BebopClient stream method

tamaralipows:


src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,395 @@
...
+
+                                    let timestamp = price_data.last_update_ts as u64;
+                                    let msg = StateSyncMessage::<TimestampHeader> {
+                                        header: TimestampHeader { timestamp },
+                                        snapshots: snapshot,
+                                        deltas: None,
+                                        removed_components: removed_components.into_iter().map(|id| (id, Default::default())).collect(),
+                                    };
+
+                                    yield (component_id, msg);
```
dianacarvalho1, line 251: wait.. you are emitting a message per pair right? 🤔 We wanted a new message with all pairs info (removed included)

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,395 @@
...
+
+        (bid_tvl + ask_tvl) / 2.0
+    }
+
+    fn create_component_with_state(
+        &self,
+        pair: &str,
+        price_data: &BebopPriceData,
+    ) -> ComponentWithState {
+        let id = format!("bebop_{}", pair.replace("/", "_"));
```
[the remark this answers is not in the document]
tamaralipows: it doesn't seem so? since the json would have duplicate keys which is impossible?

src/rfq/protocols/bebop/client.rs
tamaralipows: ah true!

src/rfq/protocols/bebop/client.rs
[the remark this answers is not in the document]
dianacarvalho1: huummm ok then I might be confusing stuff

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,395 @@
...
+            // TODO pass tvl instead of calculating it twice
+            component_tvl: Some(self.calculate_tvl(price_data)),
+            entrypoints: vec![],
+        }
+    }
+}
+
+#[async_trait]
+impl RFQClient for BebopClient {
+    fn stream(&self) -> BoxStream<'static, (String, StateSyncMessage<TimestampHeader>)> {
```
tamaralipows, line 173: This has to be a result?

tamaralipows:
```suggestion
    fn stream(&self) -> BoxStream<'static, Result<(String, StateSyncMessage<TimestampHeader>, ERROR>)> {
```

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,411 @@
...
+            .map(|(price, size)| price * size)
+            .sum();
+
+        let ask_tvl: f64 = price_data
+            .asks
+            .iter()
+            .map(|(price, size)| price * size)
+            .sum();
+
+        (bid_tvl + ask_tvl) / 2.0
```
tamaralipows, line 89: Only not sure if I should sum these tvls or avg them

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,429 @@
...
+            entrypoints: vec![],
+        }
+    }
+}
+
+#[async_trait]
+impl RFQClient for BebopClient {
+    fn stream(
+        &self,
+    ) -> BoxStream<'static, Result<(String, StateSyncMessage<TimestampHeader>), RFQError>> {
```
tamaralipows, line 141: A bit confused, if the StateSyncMessage contains ALL token pairs removed + changed, what is the String here? I guess not a component id? :/ @dianacarvalho1

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,395 @@
...
+    fn calculate_tvl(&self, price_data: &BebopPriceData) -> f64 {
+        // TODO this only works if the quote token is the same for each token pair - which I think
+        //  is not the case. Should we get token prices at this point to translate everything to
+        //  ETH or USD?
+
+        let bid_tvl: f64 = price_data
+            .bids
+            .iter()
+            .take(5) // Consider top 5 levels
+            .map(|(price, size)| price * size)
```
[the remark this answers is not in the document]
dianacarvalho1: ah yeah, you are right! sorry 😅 I clearly confused myself . yes this is correct ✅ (minus the filtering for the first 5)

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,510 @@
...
+    let mut remaining_tokens = token_amount;
+    let mut total_usdc = 0.0;
+
+    for (price, usdc_size) in price_levels.iter() {
+        if remaining_tokens <= 0.0 {
+            break;
+        }
+
+        // Convert USDC size to token quantity available at this level
+        let tokens_available = usdc_size / price;
```
dianacarvalho1, line 92: I don't understand this 🤔 why is this necessary?

dianacarvalho1: I think it should be only the usdc_size for the tokens available

dianacarvalho1: aaaahhh because you want the other token ok ok. jeezz this is confusing

src/rfq/protocols/bebop/client.rs
dianacarvalho1: I think this ok.. this only a rough estimation. I think we don't need to be super accurate here

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,510 @@
...
+/// - `price_data`: The price data containing bids and asks
+///
+/// # Returns
+/// The USDC amount at mid price, or None if insufficient liquidity on either side
+fn get_mid_usdc_price(token_amount: f64, price_data: &BebopPriceData) -> Option<f64> {
+    let sell_usdc = calculate_usdc_amount(token_amount, price_data, true)?;
+    let buy_usdc = calculate_usdc_amount(token_amount, price_data, false)?;
+
+    // Return average (mid price)
+    Some((sell_usdc + buy_usdc) / 2.0)
```
dianacarvalho1, line 63: what happens here if one of them is None?

src/rfq/protocols/bebop/client.rs
dianacarvalho1: it's just "bebop"

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,510 @@
...
+                                                continue;
+                                            };
+
+                                            let component_with_state = client.create_component_with_state(component_id.clone(), tokens, &price_data, tvl);
+                                            new_components.insert(component_id, component_with_state);
+
+                                            // Track the latest timestamp across all pairs
+                                            let timestamp = price_data.last_update_ts as u64;
+                                            if timestamp > latest_timestamp {
+                                                latest_timestamp = timestamp;
```
dianacarvalho1, line 286: why?

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,510 @@
...
+
+                                    let msg = StateSyncMessage::<TimestampHeader> {
+                                        header: TimestampHeader { timestamp: latest_timestamp },
+                                        snapshots: snapshot,
+                                        deltas: None, // Deltas are always None - all the changes are absolute
+                                        removed_components: removed_components.into_iter().map(|id| (id, Default::default())).collect(),
+                                    };
+
+                                    // Yield one message containing all updated pairs
+                                    yield Ok(("bebop_all_pairs".to_string(), msg));
```
dianacarvalho1, line 315: just "bebop" is enough. it's just the name of the RFQ

dianacarvalho1 (commented):
This is loooking veeeryy gooodd 💪🏼 A few more comments!

src/rfq/protocols/bebop/client.rs
tamaralipows: Because the timestamp differers per pair... and we can only have one timestamp in the `StateSyncMessage`? Or where do you suggest getting this timestamp? Just taking the current timestamp?

dianacarvalho1: yeah actually I think the current timestamp is ok. it's the timestamp we receive the data

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,510 @@
...
+    tycho_client::feed::synchronizer::{ComponentWithState, Snapshot, StateSyncMessage},
+    tycho_common::dto::{ProtocolComponent, ResponseProtocolState},
+    tycho_core::dto::Chain,
+};
+
+type BebopPriceMessage = HashMap<String, BebopPriceData>;
+
+#[derive(Debug, Serialize, Deserialize)]
+struct BebopPriceData {
+    last_update_ts: f64,
```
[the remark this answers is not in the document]
dianacarvalho1: oooohh do we need such precision?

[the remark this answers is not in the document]
tamaralipows: do you think it's worth to loop over all of them and remove the milliseconds? that would actually be extra work and potentially require a second object because this one I'm using to "direct decode" the json :/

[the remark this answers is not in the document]
tamaralipows: FYI

```
2025-07-17T17:10:55.994574Z ERROR tycho_simulation::rfq::protocols::bebop::client: Failed to parse websocket message: invalid type: floating point `1752772252.696952`, expected u64 at line 1 column 125

Stream error: RFQ parsing error: Failed to parse message: invalid type: floating point `1752772252.696952`, expected u64 at line 1 column 125
thread 'rfq::protocols::bebop::client::tests::test_bebop_websocket_connection' panicked at src/rfq/protocols/bebop/client.rs:495:25:
```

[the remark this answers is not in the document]
dianacarvalho1: ah ok then no. let's leave it

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,592 @@
...
+    },
+    tycho_client::feed::synchronizer::{ComponentWithState, Snapshot, StateSyncMessage},
+    tycho_common::dto::{ProtocolComponent, ResponseProtocolState},
+    tycho_core::dto::Chain,
+};
+
+type BebopPriceMessage = HashMap<String, BebopPriceData>;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct BebopPriceData {
```
dianacarvalho1, line 25: can we move this to its' own file? like models or price_data?

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,592 @@
...
+            .sum();
+
+        let mut total_tvl = (bid_tvl + ask_tvl) / 2.0;
+
+        // If quote price data is provided, we need to normalize the tvl
+        if let Some(quote_data) = quote_price_data {
+            if let Some(quote_price_usd) = quote_data.get_mid_usd_price(total_tvl) {
+                total_tvl *= quote_price_usd;
+            } else {
+                // Quote token has no TVL is USD - return 0
```
dianacarvalho1, line 63: in USD*

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,592 @@
...
+    ///
+    /// # Parameters
+    /// - `token_amount`: The amount of tokens to price
+    /// - `price_data`: The price data containing bids and asks
+    ///
+    /// # Returns
+    /// The USDC amount at mid price, or None if insufficient liquidity on either side
+    fn get_mid_usd_price(&self, token_amount: f64) -> Option<f64> {
+        let sell_usd = self.calculate_usd_amount(token_amount, true)? / token_amount;
+        let buy_usd = self.calculate_usd_amount(token_amount, false)? / token_amount;
```
dianacarvalho1, line 80: the naming is incorrect no? the first one is buy_usd and the second sell_usd. the sell flag in calculate_usd_amount if for the base token

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,592 @@
...
+                                            let mut quote_price_data: Option<BebopPriceData> = None;
+                                            // The quote token is not one of the approved quote tokens
+                                            // Get the price, so we can normalize our TVL calculation
+                                            if !client.quote_tokens.contains(token1) {
+                                                for quote_token in &client.quote_tokens {
+                                                    let quote_pair_name = format!("{}/{}", token1, quote_token);
+                                                    if let Some(data) = price_data_map.get(&quote_pair_name) {
+                                                        quote_price_data = Some(data.clone());
+                                                        break;
+                                                    };
```
dianacarvalho1, line 308: what if this is the first iteration and we don't have all possible pairs in the price_data_map?

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,592 @@
...
+
+    /// Gets the mid price of the given token with USDC as the quote token
+    ///
+    /// # Parameters
+    /// - `token_amount`: The amount of tokens to price
+    /// - `price_data`: The price data containing bids and asks
+    ///
+    /// # Returns
+    /// The USDC amount at mid price, or None if insufficient liquidity on either side
+    fn get_mid_usd_price(&self, token_amount: f64) -> Option<f64> {
```
dianacarvalho1, line 78: I find it strange that we mention usd but we don't even check if it is usd. This method and `calculate_usd_amount` are general, they are not specific to usd. We are just relying on self price levels to have a certain quote token.

dianacarvalho1 (commented):
Thank you @tamaralipows 🙏🏼 a few more questions. Tomorrow morning with a fresh brain I will review again! But I think everything is veeery correct now ✅ we are close 🤞🏼

src/rfq/protocols/bebop/client.rs
tamaralipows: then it will go into the second iteration. It won't break since we don't find the pair we need.

Or I don't know what you are asking?

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,418 @@
...
+                                        deltas: None, // Deltas are always None - all the changes are absolute
+                                        removed_components: removed_components.into_iter().map(|id| (id, Default::default())).collect(),
+                                    };
+
+                                    // Yield one message containing all updated pairs
+                                    yield Ok(("bebop".to_string(), msg));
+                                },
+                                Err(e) => {
+                                    tracing::error!("Failed to parse websocket message: {}", e);
+                                    yield Err(RFQError::ParsingError(format!("Failed to parse message: {e}")));
```
tamaralipows, line 269: @dianacarvalho1 An example of when we will yield an error and then reconnect and continue to send messages. Perhaps I should just not make it yield this error if it doesn't serve us - I can remove this tomorrow. I guess a tracing error is enough.

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,418 @@
...
+    format!("{}/{}", pair.0, pair.1)
+}
+
+#[derive(Clone)]
+pub struct BebopClient {
+    chain: Chain,
+    url: String,
+    // Pairs that we want prices for
+    pairs: HashSet<String>,
+    // Min tvl value.
```
dianacarvalho1, line 37: can you write that the TVL is in USD please?

src/rfq/protocols/bebop/client.rs
dianacarvalho1: ah no, after my latest review I withdraw this comment! AG

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,418 @@
...
+
+        // Store all bids and asks as JSON strings, since we cannot store arrays
+        if !price_data.bids.is_empty() {
+            let bids_json = serde_json::to_string(&price_data.bids).unwrap_or_default();
+            attributes.insert("bids".to_string(), bids_json.as_bytes().to_vec().into());
+        }
+        if !price_data.asks.is_empty() {
+            let asks_json = serde_json::to_string(&price_data.asks).unwrap_or_default();
+            attributes.insert("asks".to_string(), asks_json.as_bytes().to_vec().into());
+        }
```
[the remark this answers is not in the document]
tamaralipows: what do you mean by "if the key is present at all"? The decoding will fail if it's not present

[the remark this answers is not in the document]
dianacarvalho1: meh no, its fine

src/rfq/protocols/bebop/client.rs
```diff
@@ -0,0 +1,436 @@
...
+
+type BebopPriceMessage = HashMap<String, BebopPriceData>;
+
+fn pair_to_bebop_format(pair: &(String, String)) -> Result<String, RFQError> {
+    let token0 = Address::from_str(&pair.0)
+        .map_err(|_| RFQError::ParsingError(format!("Invalid token address: {}.", pair.0)))?;
+    let token1 = Address::from_str(&pair.1)
+        .map_err(|_| RFQError::ParsingError(format!("Invalid token address: {}.", pair.1)))?;
+    Ok(format!("{token0}/{token1}"))
+}
```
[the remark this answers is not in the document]
tamaralipows: Yeah I hate it but it works :(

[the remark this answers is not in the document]
tamaralipows: Let me add a comment actually that this is for checksumming

PR #275, opened 2025-07-22
fix: correctly insert balances updates

tvinagre:
This commit fixes a bug where we were completely replacing the pool state.balances on updates. For >2 tokens pool it was leading to some balances being lost. Specifically, if the state.balances was holding balances for 3 tokens and the update was only affecting 2 of them, we were deleting the 3rd token's balance from the pool state. We are now just inserting the new balances but we keep the unchanged ones.

### DIFF
--- a/src/evm/protocol/vm/state.rs
+++ b/src/evm/protocol/vm/state.rs
@@ -182,10 +182,12 @@ where
                 {
                     let sell_token_address = bytes_to_address(sell_token_address)?;
                     let buy_token_address = bytes_to_address(buy_token_address)?;
+
                     let overwrites = Some(self.get_overwrites(
                         vec![sell_token_address, buy_token_address],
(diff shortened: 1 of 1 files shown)


src/evm/protocol/vm/state.rs
```diff
@@ -357,17 +360,17 @@ where
...
+                // Prevents errors when delta balance changes do not affect all the pool tokens.
+                for (token, bal) in bals {
+                    let addr = bytes_to_address(token).map_err(|_| {
+                        SimulationError::FatalError(format!(
+                            "Invalid token address in balance update: {token:?}"
+                        ))
+                    })?;
+                    self.balances
+                        .insert(addr, U256::from_be_slice(bal));
+                }
```
zizou0x, line 373: Nice to see this back 🔥👀

zizou0x (approved):
LGTM

PR #277, opened 2025-07-22
test(Bebop): test websocket reconnection for client

tamaralipows:
- Mocks the real websocket so no internet connection is required


tamaralipows:
> uuuhhh very nice!! :clap::skin-tone-3: you async ninja now :ninja: the only thing I would change would be to make the test faster and not wait the whole 5 seconds. Can we cut this down?

@Diana The test only waits 5 seconds as a failsafe if something is broken. In reality it stops after receiving 2 messages - takes around 60ms.

PR #278, opened 2025-07-23
feat(Bebop): Quickstart example

tamaralipows:
- This doesn't yet include execution


examples/rfq_quickstart/main.rs
```diff
@@ -0,0 +1,201 @@
...
+    println!("Connecting to Bebop RFQ WebSocket...");
+    let bebop_client = BebopClient::new(
+        chain.into(), // Convert from models::Chain to dto::Chain
+        bebop_pairs,
+        cli.tvl_threshold,
+        bebop_ws_user,
+        bebop_ws_key,
+        bebop_quote_tokens,
+    )
+    .expect("Failed to create Bebop client");
```
dianacarvalho1, line 141: I'm thinking that we should make a builder for BebopClient 🤔 
This way we can have the defaults for the quote tokens for example (I think it's weird that the user needs to choose these)
Also the chain in the BebopClient should be the models one, not the dto one

dianacarvalho1: and what if they don't pass any `bebop_pairs`? 🤔 could that just mean that they want all?

examples/rfq_quickstart/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    let (tx, mut rx) = mpsc::channel::<Update>(100);
```
dianacarvalho1, line 143: this only gets 100 messages at a time right? why did you choose 100?

examples/rfq_quickstart/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    println!("Connected to Bebop RFQ! Streaming live quotes...\n");
+
+    // Start the RFQ stream in a background task
+    tokio::spawn(rfq_stream_builder.build(tx));
+
+    // Stream quotes from RFQ stream
+    while let Some(update) = rx.recv().await {
+        println!(
+            "Received RFQ quote with {} new pairs for block/timestamp {}",
```
dianacarvalho1, line 158: it's not a quote yet. these are just indicative prices

dianacarvalho1: or price levels

examples/rfq_quickstart/main.rs
```diff
@@ -0,0 +1,201 @@
...
+                    .is_subset(&HashSet::from_iter(tokens.iter()))
+                {
+                    // Try to calculate amount out using the state
+                    if let Ok(amount_out_result) =
+                        state.get_amount_out(amount_in.clone(), &sell_token, &buy_token)
+                    {
+                        let amount_out = amount_out_result.amount;
+
+                        println!(
+                            "Quote from {}: {} {} -> {} {}",
```
dianacarvalho1, line 179: same here

examples/rfq_quickstart/main.rs
```diff
@@ -0,0 +1,201 @@
...
+                            buy_token.symbol
+                        );
+                    }
+                }
+            } else {
+                println!("No matching pair found in update.");
+            }
+        }
+
+        println!("\nWaiting for more quotes... (Press Ctrl+C to exit)");
```
dianacarvalho1, line 193: and here

examples/rfq_quickstart/Readme.md
```diff
@@ -0,0 +1,43 @@
...
+- **Credentials**: Contact Bebop to obtain WebSocket API credentials for accessing live market maker quotes
+
+## What you'll see
+
+The example will:
+1. Connect to Bebop's WebSocket API using your credentials
+2. Stream live price quotes from market makers for your specified token pair
+3. Display the best available quotes with pricing information
+
+See [here](https://docs.propellerheads.xyz/tycho/for-solvers/tycho-quickstart) a complete guide on how to run the
```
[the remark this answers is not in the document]
tamaralipows: Yeah I guess so!

src/rfq/protocols/bebop/client.rs
```diff
@@ -30,12 +30,16 @@ use crate::{
 
 type BebopPriceMessage = HashMap<String, BebopPriceData>;
 
+fn checksum(address: &String) -> Result<String, RFQError> {
+    let checksum = Address::from_str(address)
+        .map_err(|_| RFQError::ParsingError(format!("Invalid token address: {address}.")))?;
+    Ok(checksum.to_checksum(None))
```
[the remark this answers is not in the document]
tamaralipows: No :( `parse_checksummed` assumes that the string is checksummed
```
Failed to create Bebop client: ParsingError("Invalid token address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2.")
```

examples/rfq_quickstart/main.rs
tamaralipows: arbitrarily :/  do you think it should be larger

examples/rfq_quickstart/main.rs
tamaralipows: (Implemented the builder pattern)

examples/rfq_quickstart/main.rs
```diff
@@ -70,8 +70,7 @@ async fn main() {
 
     let cli = Cli::parse().with_defaults();
 
-    let chain = Chain::from_str(&cli.chain)
-        .unwrap_or_else(|_| panic!("Unknown chain {chain}", chain = cli.chain));
+    let chain = cli.chain;
```
dianacarvalho1, line 74: niceee! I wasn't sure this would work tbh

examples/price_printer/main.rs
```diff
@@ -1,11 +1,14 @@
 extern crate tycho_simulation;
+#[path = "../shared/utils.rs"]
```
dianacarvalho1, line 2: aff this is a bit sucky. what about putting this in src/utils.rs ? `load_all_tokens` is there

dianacarvalho1 (approved):
Thank you @tamaralipows 🙏🏼 This looks really good! I only have really small comments now!

examples/rfq_quickstart/main.rs
tamaralipows: Yeah @dianacarvalho1

src/rfq/protocols/bebop/client.rs
```diff
@@ -83,28 +83,28 @@ pub struct BebopClient {
...
-        }
         let mut quote_tokens_checksummed: HashSet<String> = HashSet::new();
+        let mut tokens_checksummed: HashSet<String> = HashSet::new();
 
         for token in quote_tokens.iter() {
             quote_tokens_checksummed.insert(checksum(token)?.to_string());
         }
+        for token in tokens.iter() {
+            tokens_checksummed.insert(checksum(token)?.to_string());
+        }
```
dianacarvalho1, line 102: why don't we force the user to insert the token addresses as bytes already? then we don't need to worry about this

PR #282, opened 2025-07-25
feat(Hashflow): Hashflow price levels API client

tamaralipows:
- Fetches mms from Hashflow API
- Passes them to the price levels API
- Normalizes all prices against the approved quote tokens (only to filter for TVL).
- Returns components for each token pair and mm.

### DIFF
--- a/src/rfq/protocols/hashflow/client.rs
+++ b/src/rfq/protocols/hashflow/client.rs
@@ -44,7 +44,6 @@ pub struct HashflowClient {
     source: String,
     // Quote tokens to normalize to for TVL purposes. Should have the same prices.
     quote_tokens: HashSet<Bytes>,
-    market_makers: Vec<String>,
     poll_time: u64,
 }
 
@@ -69,7 +68,6 @@ impl HashflowClient {
             auth_key,
             source: auth_user,
             quote_tokens,
-            market_makers: Vec::new(), // Will be populated during first fetch
             poll_time,
         })
     }
@@ -147,7 +145,7 @@ impl HashflowClient {
         }
     }
 
-    async fn fetch_market_makers(&mut self) -> Result<(), RFQError> {
+    async fn fetch_market_makers(&mut self) -> Result<Vec<String>, RFQError> {
         let query_params = vec![
             ("source", self.source.clone()),
             ("baseChainType", "evm".to_string()),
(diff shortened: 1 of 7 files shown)


src/rfq/protocols/hashflow/client.rs
```diff
@@ -0,0 +1,535 @@
...
+        }
+
+        // Try to find the price of the quote token in one of the approved quote tokens
+        // for normalization.
+        for approved_quote_token in &self.quote_tokens {
+            for (_mm, mm_levels_inner) in levels_by_mm.iter() {
+                for quote_mm_level in mm_levels_inner {
+                    // Check for direct pair: quote_token/approved_quote_token
+                    if quote_mm_level.pair.base_token == quote_token &&
+                        quote_mm_level.pair.quote_token == *approved_quote_token
```
tamaralipows, line 97: Since this is just an estimate I'm just using the first levels I find for a price estimate regardless of if there are other mms available also offering prices. Let me know if you think it's worth to average them.

src/rfq/protocols/hashflow/models.rs
```diff
@@ -0,0 +1,199 @@
...
+fn deserialize_string_to_checksummed_bytes<'de, D>(deserializer: D) -> Result<Bytes, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    let address = Address::from_str(&s).map_err(serde::de::Error::custom)?;
+    let checksum = address.to_checksum(None);
+    let checksum_bytes = Bytes::from_str(&checksum).map_err(serde::de::Error::custom)?;
+    Ok(checksum_bytes)
+}
```
tamaralipows, line 37: Was actually a lifesaver to discover this tool for custom deserialization. Can immediately checksum and transform to Bytes straight from API output :) 

Not sure the error handling is the best here and if I can just right away return an RFQError, I will double-check it quickly tomorrow.

src/rfq/protocols/hashflow/client.rs
```diff
@@ -0,0 +1,535 @@
...
+            id: component_id.clone(),
+            protocol_system: "rfq:hashflow".to_string(),
+            protocol_type_name: "hashflow_pool".to_string(),
+            chain: self.chain.into(),
+            tokens,
+            contract_ids: vec![], // empty for RFQ
+            static_attributes: Default::default(),
+            change: Default::default(),
+            creation_tx: Default::default(),
+            created_at: Default::default(),
```
dianacarvalho1, line 129: you can just do `..Default::default()` for these 4 lines

src/rfq/protocols/hashflow/client.rs
```diff
@@ -0,0 +1,535 @@
...
+        &self,
+    ) -> BoxStream<'static, Result<(String, StateSyncMessage<TimestampHeader>), RFQError>> {
+        let mut client = self.clone();
+
+        Box::pin(async_stream::stream! {
+            let mut current_components: HashMap<String, ComponentWithState> = HashMap::new();
+            let mut ticker = interval(Duration::from_secs(10));
+            let mut market_makers_fetched = false;
+
+            info!("Starting Hashflow price levels polling every 10 seconds");
```
dianacarvalho1, line 259: 10 seconds is too long I think. maybe we can make this value a parameter?

dianacarvalho1 (commented):
Thank you so much @tamaralipows ! 🙏🏼 This already looks great!

How does the Hashflow price levels look like? a lot of weird quote tokens?
Also, should we check the available mms periodically? like every 1hr or so.. what happens if we request price levels for a mm that is not available anymore?

dianacarvalho1 (approved):
Veeery gooodd! 👏🏼  thank you! I only have one request but I won't insist on it if you don't wanna (just nitpicking a but)

PR #284, opened 2025-07-28
feat(Bebop): Use protobuf for ws connection

tamaralipows:


dianacarvalho1 (approved):
aahh this is actually nicer I would say! Thank you 🙏🏼 It looks good to me!

PR #288, opened 2025-08-01
feat(Hooks): Implement get_amount_ranges

tamaralipows:


src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -223,23 +228,85 @@ where
...
+                    args,
+                    0,                // block number (not used to get limits)
+                    None,             // timestamp
+                    None,             // overwrites
+                    None,             // caller
+                    U256::from(0u64), // value
+                    None,             // transient_storage
+                )?;
+
+                println!("Return value {:?}", res.clone());
```
dianacarvalho1, line 269: lost print

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -223,23 +228,85 @@ where
...
+
+                let decoded: (U256, U256) = (
+                    U256::from_be_slice(&res.return_value[..32]),
+                    U256::from_be_slice(&res.return_value[32..64]),
+                );
```
dianacarvalho1, line 274: I would use abi_decode here. But I won't insist in a change!

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -223,23 +228,85 @@ where
...
         &self,
-        _token_in: Bytes,
-        _token_out: Bytes,
+        token_in: Bytes,
+        token_out: Bytes,
     ) -> Result<AmountRanges, SimulationError> {
-        Err(SimulationError::RecoverableError(
-            "get_amount_ranges is not implemented for GenericVMHookHandler".to_string(),
-        ))
+        match &self.limits_entrypoint {
```
dianacarvalho1, line 234: `limits_entrypoint` is just an Option.. why do you prefer to use `match` over `if let Some(..)=`?

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -223,23 +228,85 @@ where
...
         _balances: &Balances,
     ) -> Result<(), TransitionError<String>> {
-        Err(TransitionError::SimulationError(SimulationError::RecoverableError(
-            "delta_transition is not implemented for GenericVMHookHandler".to_string(),
-        )))
+        if let Some(limits_entrypoint_bytes) = delta
+            .updated_attributes
+            .get("limits_entrypoint")
+        {
+            // TODO double check how exactly this is encoded
```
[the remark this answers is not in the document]
dianacarvalho1: should be encoded the same way here and in the snapshot I think

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -557,4 +627,97 @@ mod tests {
...
+        //
+        // // SPDX-License-Identifier: UNLICENSED
+        // pragma solidity ^0.8.26;
+        //
+        // contract Limits {
+        //     function getLimits(address a, address b) external pure returns (uint256, uint256) {
+        //         require(a != address(0) && b != address(0));
+        //         return (1000, 2000);
+        //     }
+        // }
```
dianacarvalho1, line 669: aaaahhhhhh sneaky.. I think this is ok to have as a test. But can we add a test for the real get_limits from Euler please?

dianacarvalho1 (commented):
Thank you @tamaralipows ! It looks good!! I only have small comments and a big request 😅

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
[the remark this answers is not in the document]
dianacarvalho1: [here](https://github.com/propeller-heads/tycho-simulation/pull/288/files#r2253540492) you are also decoding it from the snapshot! Why are you assuming they are encoded differently?

dianacarvalho1 (approved):
aahh very niceeee ✨ thank yooouu 🙏🏼 
so in the end, the problem with the Euler test was just using borked contracts? 😬

PR #289, opened 2025-08-05
feat: implement get_amount_out for HashflowState

adrianbenavides:


src/rfq/protocols/hashflow/state.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    fn get_amount_out(
+        &self,
+        amount_in: BigUint,
+        token_in: &Token,
+        token_out: &Token,
+    ) -> Result<GetAmountOutResult, SimulationError> {
+        // The current levels are only valid for the base/quote pair.
+        if !(token_in == &self.base_token && token_out == &self.quote_token) {
+            return Err(SimulationError::RecoverableError(format!(
```
dianacarvalho1, line 59: should be an InvalidInput in this case

dianacarvalho1 (commented):
small small comment only!

dianacarvalho1 (approved):
Thank you! 🙏🏼

PR #291, opened 2025-08-05
feat: get_limits for USV4 if no liquidity

tamaralipows:
Please refer to the [task](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?jql=&selectedIssue=ENG-4740).

`find_max_amount` uses exponential search: https://en.wikipedia.org/wiki/Exponential_search

Exponential search is likely much faster than binary search for our use case:
 - `get_amount_out(I256::MAX)` will almost always fail, so this will waste time checking values unrealistically high.
 - If you were to start binary search from 1 to 10^76, you'd need hundreds of iterations.

### DIFF
--- a/src/evm/protocol/uniswap_v4/state.rs
+++ b/src/evm/protocol/uniswap_v4/state.rs
@@ -545,7 +545,7 @@ impl ProtocolSim for UniswapV4State {
                 match hook.get_amount_ranges(token_in.clone(), token_out.clone()) {
                     Ok(amount_ranges) => {
                         return Ok((
-                            BigUint::from(0u8),
+                            u256_to_biguint(amount_ranges.amount_in_range.1),
                             u256_to_biguint(amount_ranges.amount_out_range.1),
                         ))
                     }
--- a/src/evm/protocol/uniswap_v4/state.rs
+++ b/src/evm/protocol/uniswap_v4/state.rs
@@ -292,12 +292,7 @@ impl UniswapV4State {
         let token_out_obj =
             Token::new(&token_out, "TOKEN_OUT", 18, 0, &[Some(10_000)], Default::default(), 100);
 
-        // Find limit for token_in -> token_out
-        let max_amount_in = self.find_max_amount(&token_in_obj, &token_out_obj)?;
-        // Find limit for token_out -> token_in
-        let max_amount_out = self.find_max_amount(&token_out_obj, &token_in_obj)?;
-
-        Ok((max_amount_in, max_amount_out))
+        self.find_max_amount(&token_in_obj, &token_out_obj)
     }
 
     /// Finds max amount by performing exponential search.
@@ -308,32 +303,36 @@ impl UniswapV4State {
     /// - If you were to start binary search from 1 to 10^76, you'd need hundreds of iterations.
     ///
     /// More about exponential search: https://en.wikipedia.org/wiki/Exponential_search
+    ///
+    /// # Returns
+    ///
+    /// Returns a tuple containing the max amount in and max amount out respectively.
     fn find_max_amount(
         &self,
         token_in: &Token,
         token_out: &Token,
-    ) -> Result<BigUint, SimulationError> {
+    ) -> Result<(BigUint, BigUint), SimulationError> {
         let mut low = BigUint::from(1u64);
 
         // The max you can swap on a USV4 is I256::MAX is 5.7e76, since input amount is I256.
         // So start with something much smaller to search for a reasonable upper bound.
         let mut high = BigUint::from(10u64).pow(18); // 1 ether in wei
-        let mut last_successful = BigUint::from(1u64);
+        let mut last_successful_amount_in = BigUint::from(1u64);
+        let mut last_successful_amount_out = BigUint::from(0u64);
 
         // First, find an upper bound where the swap fails using exponential search.
-        while self
-            .get_amount_out(high.clone(), token_in, token_out)
-            .is_ok()
-        {
+        // Save and return both the amount in and amount out.
+        while let Ok(result) = self.get_amount_out(high.clone(), token_in, token_out) {
             // We haven't found the upper bound yet, increase the attempted upper bound
             // by order of magnitude and store the last success as the lower bound.
-            low = last_successful;
-            last_successful = high.clone();
+            low = last_successful_amount_in.clone();
+            last_successful_amount_in = high.clone();
+            last_successful_amount_out = result.amount;
             high *= BigUint::from(10u64);
 
             // Stop if we're getting too large for I256 (about 10^75)
             if high > BigUint::from(10u64).pow(75) {
-                return Ok(last_successful);
+                return Ok((last_successful_amount_in, last_successful_amount_out));
             }
         }
 
@@ -342,8 +341,9 @@ impl UniswapV4State {
             let mid = (&low + &high) / BigUint::from(2u64);
 
             match self.get_amount_out(mid.clone(), token_in, token_out) {
-                Ok(_) => {
-                    last_successful = mid.clone();
+                Ok(result) => {
+                    last_successful_amount_in = mid.clone();
+                    last_successful_amount_out = result.amount;
                     low = mid;
                 }
                 Err(_) => {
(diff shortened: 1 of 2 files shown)


src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -279,6 +279,79 @@ impl UniswapV4State {
...
+        let mut low = BigUint::from(1u64);
+
+        // The max you can swap on a USV4 is I256::MAX is 5.7e76, since input amount is I256.
+        // So start with something much smaller to search for a reasonable upper bound.
+        let mut high = BigUint::from(10u64).pow(18); // 1 ether in wei
+        let mut last_successful = BigUint::from(1u64);
+
+        // First, find an upper bound where the swap fails using exponential search.
+        while self
+            .get_amount_out(high.clone(), token_in, token_out)
```
adrianbenavides, line 318: Out of the scope of this PR, but the `amount_in` in `get_amount_out` could be passed as a reference right? That would remove a lot of biguint clonings. I'm not very familiar with the biguint type, but it seems to be pretty expensive to clone if the number is high enough.

I would also move the initialization of constant biguints outside of the loops. Even if they are small numbers, they are still allocating/deallocating vec's in each iteration.

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -279,6 +279,79 @@ impl UniswapV4State {
...
+
+        // Find limit for token_in -> token_out
+        let max_amount_in = self.find_max_amount(&token_in_obj, &token_out_obj)?;
+        // Find limit for token_out -> token_in
+        let max_amount_out = self.find_max_amount(&token_out_obj, &token_in_obj)?;
+
+        Ok((max_amount_in, max_amount_out))
+    }
+
+    /// Finds max amount by performing exponential search.
```
adrianbenavides, line 303: Could you add more context here on why exponential search is being used? I think the paragraph you have in the PR description would be very helpful here.

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -279,6 +279,79 @@ impl UniswapV4State {
...
+            Token::new(&token_in, "TOKEN_IN", 18, 0, &[Some(10_000)], Default::default(), 100);
+        let token_out_obj =
+            Token::new(&token_out, "TOKEN_OUT", 18, 0, &[Some(10_000)], Default::default(), 100);
+
+        // Find limit for token_in -> token_out
+        let max_amount_in = self.find_max_amount(&token_in_obj, &token_out_obj)?;
+        // Find limit for token_out -> token_in
+        let max_amount_out = self.find_max_amount(&token_out_obj, &token_in_obj)?;
+
+        Ok((max_amount_in, max_amount_out))
```
dianacarvalho1, line 300: I think this is unnecessary 🤔 afaiu the `get_limits` returns the max amount in and the max amount out possible. and the max amount is the one used to get the max amount out. 
Here you are finding the two max amount outs (one for each token). The user should be the one to do this by calling the `get_limits` with different token_in/token_out

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -279,6 +279,79 @@ impl UniswapV4State {
...
+        &self,
+        token_in: Bytes,
+        token_out: Bytes,
+    ) -> Result<(BigUint, BigUint), SimulationError> {
+        // Create dummy token objects with proper addresses. This is fine since `get_amount_out`
+        // only uses the token addresses.
+        let token_in_obj =
+            Token::new(&token_in, "TOKEN_IN", 18, 0, &[Some(10_000)], Default::default(), 100);
+        let token_out_obj =
+            Token::new(&token_out, "TOKEN_OUT", 18, 0, &[Some(10_000)], Default::default(), 100);
```
dianacarvalho1, line 293: oh this soooo sneaky. I like it

src/evm/protocol/uniswap_v4/state.rs
dianacarvalho1: but thank you for these comments 🙏🏼 and please keep them coming!

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -984,4 +1082,205 @@ mod tests {
...
+            .is_ok());
+        assert!(usv4_state
+            .get_amount_out(max_amount.clone() + BigUint::from(1u64), &token_in, &token_out)
+            .is_err());
+    }
+
+    #[test]
+    fn test_find_max_amount_less_than_starting_point() {
+        // We start our search at 1e18, so test that we can properly find a max amount less than
+        // that
```
dianacarvalho1, line 1200:
```suggestion
        // We start our search at 1e18, so test that we can properly find a max amount smaller than
        // that
```

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -984,4 +1082,205 @@ mod tests {
...
+        let one_more = &max_amount + BigUint::from(1u64);
+        let should_fail = usv4_state
+            .get_amount_out(one_more, &token_in, &token_out)
+            .is_err();
+        assert!(should_fail, "Swapping max_amount + 1 should fail");
+    }
+
+    #[test]
+    fn test_find_max_amount_boundary_conditions() {
+        // Test edge cases and boundary conditions
```
dianacarvalho1, line 1246: what exactly are the edge cases and boundary conditions here?

dianacarvalho1 (commented):
Thank you so much @tamaralipows 🙏🏼 Great work! I have a few comments

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -459,17 +532,28 @@ impl ProtocolSim for UniswapV4State {
...
+                    Ok(amount_ranges) => {
+                        return Ok((
+                            BigUint::from(0u8),
+                            u256_to_biguint(amount_ranges.amount_out_range.1),
+                        ))
+                    }
+                    // Check if hook get_amount_ranges is not implemented or the limits entrypoint
+                    // is not set for this hook
+                    Err(SimulationError::RecoverableError(msg))
+                        if msg.contains("not implemented") || msg.contains("not set") =>
```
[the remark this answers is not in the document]
tamaralipows: I agree we should think more about this! I would perhaps open a separate task for this since it seems quite important to be on the same page. I would also combine this with your comments about error handling on Diana's PR

src/evm/protocol/uniswap_v4/state.rs
tamaralipows: I've done a test refactor to parametrize this actually, it's a lot cleaner now. Have a look at my latest commit!

src/evm/protocol/uniswap_v4/state.rs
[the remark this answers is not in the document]
adrianbenavides: 100% agree 👍

src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -459,17 +532,28 @@ impl ProtocolSim for UniswapV4State {
...
-                Err(SimulationError::RecoverableError(msg)) if msg.contains("not implemented") => {
-                    // Fall back to default implementation
+            // Check if pool has no liquidity & ticks -> hook manages liquidity
+            if self.liquidity == 0 && self.has_no_initialized_ticks() {
+                // If the hook has a get_amount_ranges entrypoint, call it and return (0, limits[1])
+                match hook.get_amount_ranges(token_in.clone(), token_out.clone()) {
+                    Ok(amount_ranges) => {
+                        return Ok((
+                            BigUint::from(0u8),
+                            u256_to_biguint(amount_ranges.amount_out_range.1),
```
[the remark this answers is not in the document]
dianacarvalho1: the task description is wrong.. That was just to mean to we return only the max ranges and not the min ranges. But what is described there doesn't match with actual interface, sorry 😕

src/evm/protocol/uniswap_v4/state.rs
dianacarvalho1: yes yes, that's exactly what I meant!

src/evm/protocol/uniswap_v4/state.rs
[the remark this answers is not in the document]
tamaralipows: updated!

dianacarvalho1 (approved):
merci merci 🙏🏼

PR #294, opened 2025-08-06
feat: angstrom

Will-Smith11:


Will-Smith11:
Things remaining
1) need to wire in the unlock attestation into the encoder
2) need to test

examples/quickstart/main.rs
```diff
@@ -186,6 +187,11 @@ async fn main() {
                     tvl_filter.clone(),
                     Some(uniswap_v4_pool_with_hook_filter),
                 )
+                .exchange::<UniswapV4State>(
+                    "angstrom",
+                    tvl_filter.clone(),
+                    Some(angstrom_pool_filter),
+                )
```
kayibal, line 194: I wasn't thinking of modelling angstrom as a separate exchange, instead it is just part of v4. Solvers can filter/include by hooks by modifying this filter when they subscribe.

src/evm/protocol/filters.rs
```diff
@@ -169,6 +169,26 @@ pub fn uniswap_v4_pool_with_hook_filter(component: &ComponentWithState) -> bool
...
+        if hooks.to_vec() != **ANGSTROM_ADDRESS {
+            debug!(
+                "Filtering out UniswapV4 pool {} because it is not angstrom",
+                component.component.id
+            );
+            return false;
+        }
+    }
+    true
+}
```
kayibal, line 190: Most likely we won't need this. Maybe we can have the ANGSTROM_ADDRESS constant somewhere exposed.

src/evm/protocol/uniswap_v4/angstrom/streams.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+    pub async fn send_unlock_update(&mut self, update: ConsensusDataWithBlock<Bytes>) {
+        todo!()
+    }
+
+    pub async fn on_tycho_message(&mut self, feed_message: Arc<FeedMessage>) {
+        if let Some(state_message) = feed_message.state_msgs.get("angstrom") {
+            self.tycho_header = state_message.header.clone();
+        }
+    }
```
kayibal, line 150: I think you want to query `"uniswap-v4"` here, not `"angstrom"`.

But the main issue here is that we potentially want to undo previously sent Slot0 updates in case they are not included in the block or is this absolutely impossible to happen? I am ok to ignore for now but might leave the pool in a potential "stale" state.

Will-Smith11: So slot0 updates will stream valid pool state in the case that there are no orders, because of this, the only thing that needs to be added is detecting if the slot0 stream stops sending the 100ms updates (will attempt to reconnect) and then we want to rollback to the lastest state to avoid staleness, However If slot0 stream is working fine, It will just stream the correct state itself so we don't have to worry about what the tycho-indexer is sending.

PR #295, opened 2025-08-07
feat: implement spot_price and get_limits for hashflow client

adrianbenavides:


src/rfq/protocols/hashflow/state.rs
```diff
@@ -45,7 +45,12 @@ impl ProtocolSim for HashflowState {
     }
 
     fn spot_price(&self, _base: &Token, _quote: &Token) -> Result<f64, SimulationError> {
-        todo!()
+        // Hashflow's levels are sorted by price, so the first level represents the best price.
+        self.levels
```
dianacarvalho1, line 49: also please check if the tokens match what we expected! Remember that Hashflow is _special_ and supports swaps only in one direction

src/rfq/protocols/hashflow/state.rs
```diff
@@ -102,10 +107,35 @@ impl ProtocolSim for HashflowState {
...
+            self.levels
+                .levels
+                .iter()
+                .fold((0.0, 0.0), |(sell_sum, buy_sum), level| {
+                    (sell_sum + level.quantity, buy_sum + level.quantity * level.price)
+                });
+
+        let sell_limit =
+            BigUint::from((total_sell_amount * 10_f64.pow(sell_decimals as f64)) as u128);
+        let buy_limit = BigUint::from((total_buy_amount * 10_f64.pow(buy_decimals as f64)) as u128);
```
dianacarvalho1, line 136: this looks correct ✅

dianacarvalho1 (commented):
Thank you @adrianbenavides ! 🙏🏼  I only have one concern

src/rfq/protocols/hashflow/state.rs
```diff
@@ -37,15 +37,47 @@ impl HashflowState {
...
-        todo!()
+    fn spot_price(&self, base: &Token, quote: &Token) -> Result<f64, SimulationError> {
+        self.valid_direction_guard(&base.address, &quote.address)?;
+
+        // Hashflow's levels are sorted by price, so the first level represents the best price.
+        self.levels
+            .levels
+            .first()
+            .ok_or(SimulationError::RecoverableError("No liquidity".into()))
+            .map(|level| level.price)
```
tamaralipows, line 80: I think this is fine - Bebop also sorted by price leaving the best one first! So no need to do the work twice by finding the min :)

tamaralipows (approved):
Thank you!

PR #298, opened 2025-08-12
feat(Bebop): Support AggregateOrders when requesting the quote

dianacarvalho1:
[task](https://propeller-heads.atlassian.net/jira/software/projects/ENG/boards/24?selectedIssue=ENG-4791)

### DIFF
--- a/src/rfq/protocols/bebop/client.rs
+++ b/src/rfq/protocols/bebop/client.rs
@@ -28,7 +28,9 @@ use crate::{
         client::RFQClient,
         errors::RFQError,
         models::TimestampHeader,
-        protocols::bebop::models::{BebopPriceData, BebopPricingUpdate, BebopQuoteResponse},
+        protocols::bebop::models::{
+            BebopOrderToSign, BebopPriceData, BebopPricingUpdate, BebopQuoteResponse,
+        },
     },
     tycho_client::feed::synchronizer::{ComponentWithState, Snapshot, StateSyncMessage},
     tycho_common::dto::{ProtocolComponent, ResponseProtocolState},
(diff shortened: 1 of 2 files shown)


src/rfq/protocols/bebop/client.rs
```diff
@@ -354,9 +356,11 @@ impl RFQClient for BebopClient {
                 ("expiry_type", "standard".into()),
                 ("fee", "0".into()),
                 ("is_ui", "false".into()),
+                ("source", self.ws_user.clone()),
             ])
             .header("accept", "application/json")
             .header("name", &self.ws_user)
+            .header("source-auth", &self.ws_key)
```
tamaralipows, line 363: One concern: do we assume that all users will have an auth key?

dianacarvalho1: Adding a cooldown task to handle this!

tamaralipows: Sooo actually you can... I just ran it fine like this
```
        let request = client
            .get(&url)
            .query(&[
                ("sell_tokens", sell_token),
                ("buy_tokens", buy_token),
                ("sell_amounts", sell_amount),
                ("taker_address", sender),
                ("receiver_address", receiver),
                ("approval_type", "Standard".into()),
                ("skip_validation", "true".into()),
                ("skip_taker_checks", "true".into()),
                ("gasless", "false".into()),
                ("expiry_type", "standard".into()),
                ("fee", "0".into()),
                ("is_ui", "false".into()),
                // ("source", self.ws_user.clone()),
            ])
            .header("accept", "application/json");
            // .header("name", &self.ws_user)
            // .header("source-auth", &self.ws_key)
            // .header("Authorization", &self.ws_key);

tamaralipows: It seems the api key is only necessary for getting price levels...

tamaralipows (approved):

PR #300, opened 2025-08-14
feat: Add execution to RFQ quickstart

dianacarvalho1:
- [x] Added encoding
- [x] Added execution
- [x] Made execution pass with high slippage (wait for permit2 tx before encoding)
- [x] Make execution pass with 0.25% slippage


dianacarvalho1 (commented):
I would approve if I could 😅

PR #303, opened 2025-08-18
feat: Don't pass PK as CLI command

tamaralipows:
- Passing it via the CLI is unsafe, since cargo runs a build log echoing the command (and thus the private key), which may be unintentionally saved to a log file if this is the user's setup.


dianacarvalho1 (approved):
veeeryy niceeeee ✨ me like this!

PR #304, opened 2025-08-19
fix: add missing fields when parsing hashflow quote

adrianbenavides:
The `request_binding_quote` implementation for Hashflow is missing some required fields by the [executor](https://github.com/propeller-heads/tycho-execution/blob/a01229f2ec5f7efdddf863c40c7feec11b7f18ca/foundry/src/executors/HashflowExecutor.sol#L104-L116).

### DIFF
--- a/src/rfq/protocols/hashflow/client.rs
+++ b/src/rfq/protocols/hashflow/client.rs
@@ -432,6 +432,13 @@ impl RFQClient for HashflowClient {
                     quote_attributes.insert("pool".to_string(), quote.quote_data.pool);
                     if let Some(external_account) = quote.quote_data.external_account {
                         quote_attributes.insert("external_account".to_string(), external_account);
+                    } else {
+                        quote_attributes.insert(
+                            "external_account".to_string(),
+                            Bytes::from_str(&Address::ZERO.to_string()).map_err(|_| {
+                                RFQError::ParsingError("Failed to parse zero address".to_string())
+                            })?,
+                        );
                     }
                     quote_attributes.insert("trader".to_string(), quote.quote_data.trader);
                     quote_attributes.insert("base_token".to_string(), quote.quote_data.base_token);


src/rfq/protocols/hashflow/client.rs
```diff
@@ -433,35 +431,68 @@ impl RFQClient for HashflowClient {
 
                     let mut quote_attributes: HashMap<String, Bytes> = HashMap::new();
                     quote_attributes.insert("pool".to_string(), quote.quote_data.pool);
+                    if let Some(external_account) = quote.quote_data.external_account {
+                        quote_attributes.insert("external_account".to_string(), external_account);
```
dianacarvalho1, line 435: what exactly is the external_account in the end?

src/rfq/protocols/hashflow/client.rs
```diff
@@ -433,35 +431,68 @@ impl RFQClient for HashflowClient {
...
+                    }
                     quote_attributes.insert("trader".to_string(), quote.quote_data.trader);
+                    if let Some(effective_trader) = quote.quote_data.effective_trader {
+                        quote_attributes.insert("effective_trader".to_string(), effective_trader);
+                    }
```
dianacarvalho1, line 440: I thought we didn't need this in the end..

src/rfq/protocols/hashflow/client.rs
```diff
@@ -433,35 +431,68 @@ impl RFQClient for HashflowClient {
...
+                    quote_attributes.insert("base_token".to_string(), quote.quote_data.base_token);
```
dianacarvalho1, line 441: aaaahhh I think I understand what you are doing here.. You are adding everything here to make encoding easier.. but we already have this info in the `SignedQuote.base_token`. Do we need to have it in multiple places? I don't have a strong strong preference in this though

dianacarvalho1: same for `quote_token`, `base_token_amount` and `quote_token_amount`

src/rfq/protocols/hashflow/client.rs
```diff
@@ -433,35 +431,68 @@ impl RFQClient for HashflowClient {
...
-                    quote_attributes.insert("tx_id".to_string(), quote.quote_data.tx_id);
-                    quote_attributes.insert("signature".to_string(), quote.signature);
                     quote_attributes.insert(
                         "quote_expiry".to_string(),
                         Bytes::from(
-                            quote
-                                .quote_data
-                                .quote_expiry
-                                .to_be_bytes()
+                            biguint_to_u256(&BigUint::from(quote.quote_data.quote_expiry))
```
dianacarvalho1, line 481: would it work to define `HashflowQuoteData.quote_expiry` as a U256? because right now it is a u64. And we are converting to BigUint to convert to U256 to convert to Bytes 😬 at least the BigUint step can be removed for sure

dianacarvalho1 (commented):
Thank you @adrianbenavides ! 🙏🏼 I have a few suggestions

src/rfq/protocols/hashflow/client.rs
dianacarvalho1: aaahhhh so we are missing that bit no? to check if it is passed, and if not use the zero address

src/rfq/protocols/hashflow/client.rs
dianacarvalho1: but for our case it will always be the zero address correct? we can save a tiny bit of gas if we don't pass this and just define it in the executor

dianacarvalho1 (approved):
Thank you 🙏🏼

PR #306, opened 2025-08-19
fix: Fast-forward to latest message in quickstart

tamaralipows:
The problem this fixes:

- If you leave the quickstart running, do something else, come back, there will likely be negative slippage because now the price levels have been there for a while, and the quote will be different. The unintuitive part is, it’s not going to recover from this if you just get the next price level, since now the timestamps are delayed/ It doesn’t just give you most up to date timestamp… it gives you the timestamp a few seconds after the last one. You would need to cycle through all of them to get the most recent and thus fix the negative slippage issue.


examples/quickstart/main.rs
```diff
@@ -235,6 +235,26 @@ async fn main() {
             }
         };
 
+        // Drain any additional buffered messages to get the most recent one
```
dianacarvalho1, line 238: oh no no no no no, we can't do this in the normal quickstart!! Remember that for on chain protocols, we get deltas.. so we need to apply them all to the states, otherwise we will end up with weird states that do not correspond to the onchain state

examples/rfq_quickstart/main.rs
```diff
@@ -179,6 +179,26 @@ async fn main() {
 
     // Stream quotes from RFQ stream
     while let Some(update) = rx.recv().await {
+        // Drain any additional buffered messages to get the most recent one
```
dianacarvalho1, line 182: can we add a big warning here please? if the user wants to track removed pairs, then they can not do this because the removed pairs are relative to the previous update! 
if they assume that wtv it is in the message is the current state (like we are doing in the quickstart) then this is fine

dianacarvalho1 (commented):
Thank you @tamaralipows ! I have one concern!

PR #314, opened 2025-08-27
fix: add deserialize support to protocolComponent

quasystaty1:
protocolComponent only supports serialize at the moment.


dianacarvalho1 (approved):
Thank you! Looks good to me 👍🏼

PR #326, opened 2025-09-09
feat: Allow clearing PreCachedDB account storage

tamaralipows:
Main issue:
- This was causing a big issue in the SDK testing, since accounts were being leaked between test cases. Adding this functionality provides an easy way to clear storage on the rust side, by calling .clear() between subsequent tests.

How this manifested:
- Incorrect pool balances were being loaded in the BalancerV2SwapAdapter if tests were run altogether.
- Pool balances were correct if tests were run in isolation.
- We checked that tycho-indexer returned correct pool balances, and we checked pool balances for that block on-chain to be double-sure
- In the python test suite, which was passing, there was code which dropped the whole database between subsequent tests, which was not being duplicated in the rust suite.


src/evm/engine_db/tycho_db.rs
```diff
@@ -230,6 +230,15 @@ impl PreCachedDB {
             .as_ref()
             .map(|header| header.number)
     }
+
+    /// Clear all state from the database.
+    ///
+    /// This is useful for test isolation to ensure state doesn't leak between tests.
```
tvinagre, line 236: I think it's irrelevant to comment here that this is done to avoid leaking between tests - it's not the function's purpose and it can be used in other cases.

tvinagre (approved):

PR #333, opened 2025-09-18
fix: Update block in UniswapV4State and EVMPoolState

tamaralipows:
- the issue stemmed from UniswapV4State. It's likely not an issue yet in EVMPoolState.
- beforeSwap relies on the block timestamp in the case of Euler. We get an arithmetic over/underflow if this is wrong. For an  example of where this is problematic in the EVault:
```
uint256 deltaT = block.timestamp - vaultCache.lastInterestAccumulatorUpdate;
<=> uint256 deltaT = 1758201863 - 1758201935;
```

### DIFF
--- a/src/evm/protocol/uniswap_v4/state.rs
+++ b/src/evm/protocol/uniswap_v4/state.rs
@@ -712,6 +712,33 @@ impl ProtocolSim for UniswapV4State {
             }
         }
 
+        // Update block information if provided in the delta attributes
+        if let Some(block_number_bytes) = delta
+            .updated_attributes
+            .get("block_number")
+        {
+            self.block.number = u64::from_be_bytes(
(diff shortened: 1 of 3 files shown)


src/evm/protocol/uniswap_v4/state.rs
```diff
@@ -712,6 +712,33 @@ impl ProtocolSim for UniswapV4State {
             }
         }
 
+        // Update block information if provided in the delta attributes
+        if let Some(block_number_bytes) = delta
```
tvinagre, line 716: Is it dangerous if we only update the block_number and not the timestamp?

tvinagre: Yes, pls raise a warning if one is missing, please - so we know something is funky

PR #335, opened 2025-09-20
feat: Remove block from SimulationParams, retrieve from DB instead

tamaralipows:
- The VM always has the latest block available by default. So if we don’t override it we should have the latest one. (Not working)
- In the case of simulating in the past, SimulationDB and PreCachedDBInner both have a block: Option<BlockHeader>  attribute that is always kept up to date...
so we could use the block from this in the actual simulation, instead of passing it around inside of the SimulationParameters.
- I tried just not setting the block, but this was giving the same arithmetic overflow/underflow error. Retrieving it from the DB ran smoothly for an hour.

Tried to organize the commits as nicely as I could... though it might still be confusing, sorry. If it's too overwhelming let's have a call.

[Issue](https://propeller-heads.atlassian.net/browse/ENG-5030) (No additional info)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,48 +17,57 @@ members = ["tycho_simulation_py"]
 
 [dependencies]
 # Serialization/Deserialization
+chrono = { version = "0.4.26", features = ["serde"] }
+hex = "0.4.3"
 serde = { version = "1.0", features = ["rc"] }
 serde_json = "1.0.105"
-uuid = { version = "1.4.1", features = ["serde", "v4", "fast-rng", "macro-diagnostics"] }
-hex = "0.4.3"
-chrono = { version = "0.4.26", features = ["serde"] }
+uuid = { version = "1.4.1", features = [
+    "serde",
+    "v4",
+    "fast-rng",
+    "macro-diagnostics",
+] }
 
 # Error handling
 thiserror = "1"
 
 # Async & concurrency
-tokio = { version = "1.38.0", features = ["full"] }
 futures = "0.3.31"
+tokio = { version = "1.38.0", features = ["full"] }
 
 # Logging & Tracing
 tracing = "0.1.37"
 
 # Utility
-num-traits = "0.2.17"
 dotenv = "0.15.0"
-itertools = "0.10.5"
 enum_delegate = "0.2.0"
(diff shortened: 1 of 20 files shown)


src/utils.rs
```diff
@@ -94,7 +94,7 @@ pub async fn load_all_tokens(
 /// Get the default Tycho URL for the given chain.
 pub fn get_default_url(chain: &Chain) -> Option<String> {
     match chain {
-        Chain::Ethereum => Some("tycho-beta.propellerheads.xyz".to_string()),
+        Chain::Ethereum => Some("tycho-dev.propellerheads.xyz".to_string()),
```
dianacarvalho1, line 97: upsi

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tycho-simulation"
-version = "0.161.0"
+version = "0.160.1"
```
dianacarvalho1, line 3: why did this change here?

tamaralipows: i have changed this back like 5 times, i don't know why it keeps reverting to 1.60.1 -_-

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -639,11 +629,23 @@ mod tests {
                 "0x639d7e454339ba43da3b2288b45078405330afcc3cd7f10e6e852be9c70ac164",
             )
             .unwrap(),
-            timestamp: 1748397011,
+            timestamp: std::time::SystemTime::now()
+                // Unless we would like to overwrite the Euler contract storage,
+                // we must use the current timestamp, otherwise we will end up with an arithmetic underflow
+                // because of this line `uint256 deltaT = block.timestamp - vaultCache.lastInterestAccumulatorUpdate;`
+                // This is expected, since SimulationDB is incapable of simulating in the past
```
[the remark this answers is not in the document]
tamaralipows: I talked to @tvinagre about this yesterday and he said it's not capable of simulating in the past...

By this, I mean:

We can overwrite the `block.number` and `block.timestamp` so that it uses these things for computations when needed, but it doesn't literally use the **past state** of each contract that it touches...

PR #338, opened 2025-09-24
feat: Use custom contract to fetch Euler limits with the correct values

tamaralipows:
Adapter from related code on tycho-indexer: https://github.com/propeller-heads/tycho-indexer/pull/690

This is very hacky but it should be a temporary fix until Euler releases v2

I've hardcoded the block number (to be sure we are simulating on the same block) and tested before and after my changes, to be sure that this actually makes a difference. The result...

```
BlockEnv { number: 23437084, beneficiary: 0x0000000000000000000000000000000000000000, timestamp: 1758767447, gas_limit: 18446744073709551615, basefee: 0, difficulty: 0, prevrandao: Some(0x0000000000000000000000000000000000000000000000000000000000000000), blob_excess_gas_and_price: Some(BlobExcessGasAndPrice { excess_blob_gas: 0, blob_gasprice: 1 }) }

Before
Max in:  50575144805324
Max out: 348319231566699

After:
Max in:  50575144805324
Max out: 348319231416624
```


src/evm/protocol/uniswap_v4/hooks/hook_handler_creator.rs
```diff
@@ -75,6 +75,13 @@ impl HookHandlerCreator for GenericVMHookHandlerCreator {
             .get("limits_entrypoint")
             .and_then(|bytes| String::from_utf8(bytes.0.to_vec()).ok());
 
+        let is_euler;
+        if let Some(_hooks_data) = params.attributes.get("hook_identifier") {
```
tvinagre, line 79: Actually Euler will have `euler_v1` as hook_identifier but we can address this in a future PR

tvinagre (approved):

src/evm/protocol/uniswap_v4/hooks/generic_vm_hook_handler.rs
```diff
@@ -73,12 +79,14 @@ where
         _all_tokens: HashMap<Bytes, Token>,
         _account_balances: HashMap<Bytes, HashMap<Bytes, Bytes>>,
         limits_entrypoint: Option<String>,
+        is_euler: bool,
```
dianacarvalho1, line 82: 😂 amazing

PR #345, opened 2025-09-29
feat: Upgrade revm, foundry, solar and alloy deps

dianacarvalho1:
Fix ekubo test. The failure happens because alloy had a big upgrade and there was a slight change in how certain mathematical operations are performed (the rounding is slightly different)

### DIFF
--- a/src/evm/protocol/ekubo/test_cases.rs
+++ b/src/evm/protocol/ekubo/test_cases.rs
@@ -584,7 +584,7 @@ pub fn mev_resist() -> TestCase {
             ("tick".to_string(), TICK_INDEX_BETWEEN.to_be_bytes().into()),
         ]
         .into(),
-        swap_token0: (100_u8.into(), 87_u8.into()),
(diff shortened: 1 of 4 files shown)


src/evm/protocol/ekubo/test_cases.rs
```diff
@@ -584,7 +584,7 @@ pub fn mev_resist() -> TestCase {
             ("tick".to_string(), TICK_INDEX_BETWEEN.to_be_bytes().into()),
         ]
         .into(),
-        swap_token0: (100_u8.into(), 87_u8.into()),
```
zizou0x, line 587: Why did this change? 😮

dianacarvalho1: I wrote it in the commit message 😅 
The failure happens because alloy had a big upgrade and there was a slight change in how certain mathematical operations are performed (the rounding is slightly different)

PR #347, opened 2025-09-30
fix: make sure we don't emit any update after we skipped a decoder failure for a component

zizou0x:
This fix is very conservative. The idea is simple: if there is a decoding failure and `skip_state_decode_failures` is on, we emit the removal of the pool and make sure we never publish any new state for it (until it restarts). This will help making sure we're not serving wrong state to clients.

Every failure emits a warn so it's easy to catch them and investigate.

### DIFF
--- a/src/evm/decoder.rs
+++ b/src/evm/decoder.rs
@@ -2,7 +2,6 @@ use std::{
     collections::{hash_map::Entry, HashMap, HashSet},
     future::Future,
     pin::Pin,
-    str::FromStr,
     sync::Arc,
 };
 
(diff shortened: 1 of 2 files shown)


src/evm/decoder.rs
```diff
@@ -778,6 +775,25 @@ where
...
         // Persist the newly added/updated states
         let mut state_guard = self.state.write().await;
+
+        // Update failed components with any new ones
+        state_guard
+            .failed_components
+            .extend(msg_failed_components);
+
+        // Remove any failed components from Updates
+        // Perf: we could do it directly in the decoder logic to avoid some steps, but this logic is complex and this is more robust.
```
zizou0x, line 785: As not serving wrong state is very critical, I've decided to go for the easiest and just filter out at the end. This is not the best for performance but it's the most robust and future proof (because it's impossible to miss any update location). Let me know if you think this is terrible.

kayibal (approved):

tvinagre (approved):

PR #348, opened 2025-09-30
docs: Document stream module.

kayibal:
This informs our users exactly about the behaviour of the stream and allows them to structure their applications respectively.

### DIFF
--- a/src/evm/stream.rs
+++ b/src/evm/stream.rs
@@ -1,4 +1,102 @@
-use std::{collections::HashMap, sync::Arc};
+//! Builder for configuring a multi-protocol stream.
+//!
+//! Provides a builder for creating a multi-protocol stream that produces
+//! [`protocol::models::Update`] messages. It runs one synchronization worker per protocol
+//! and a supervisor that aggregates updates, ensuring gap‑free streaming
+//! and robust state tracking.
+//!
+//! ## Context
+//!
+//! This stream wraps [`tycho_client::stream::TychoStream`]. It decodes `FeedMessage`s
+//! into [`protocol::models::Update`]s. Internally, each protocol runs in its own
+//! synchronization worker, and a supervisor aggregates their messages per block.
+//!
+//! ### Protocol Synchronization Worker
+//! A synchronization worker runs the snapshot + delta protocol from `tycho-indexer`.
+//! - It first downloads components and their snapshots.
+//! - It then streams deltas.
+//! - It reacts to new or paused components by pulling snapshots or removing them from the active
+//!   set.
+//!
+//! Each worker emits snapshots and deltas to the supervisor.
+//!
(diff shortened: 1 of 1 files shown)


src/evm/stream.rs
```diff
@@ -91,9 +168,26 @@ impl ProtocolStreamBuilder {
...
+    /// - `filter`: Defines the set of components to include in the stream.
+    /// - `filter_fn`: Optional custom filter function for client-side filtering of components not
+    ///   expressible in `filter`.
+    ///
+    /// # Notes
+    ///
+    /// For certain protocols (e.g., `"uniswap_v4"`, `"vm:balancer_v2"`, `"vm:curve"`), omitting
+    /// `filter_fn` may cause decoding errors or incorrect results. In these cases, a proper
+    /// filter function is required to ensure correct decoding and quoting logic.
     pub fn exchange<T>(
```
dianacarvalho1, line 191: The comment says 

> Adds a specific protocol to the stream.

why is the method name exchange? 😅

src/evm/stream.rs
```diff
@@ -149,36 +277,66 @@ impl ProtocolStreamBuilder {
...
         self.decoder.set_tokens(tokens).await;
         self
     }
 
-    /// Skips state decode failures, allowing the stream to continue processing. It raises a warning
-    /// instead of panic.
+    /// Skips decoding errors for component state updates.
+    ///
+    /// Allows the stream to continue processing even if some states fail to decode,
+    /// logging a warning instead of panicking.
```
dianacarvalho1, line 313: do we recommend using this?

dianacarvalho1 (commented):
Thank you @kayibal ! Overall I like these docs! Just have a few questions..

src/evm/stream.rs
kayibal: Yes

src/evm/stream.rs
```diff
@@ -1,4 +1,99 @@
...
+//! ### Protocol Synchronization Worker
+//! A synchronization worker runs the snapshot + delta protocol from `tycho-indexer`.
+//! - It first downloads components and their snapshots.
+//! - It then streams deltas.
+//! - It reacts to new or paused components by pulling snapshots or removing them from the active
+//!   set.
+//!
+//! Each worker emits snapshots and deltas to the supervisor.
+//!
+//! ### Stream Supervisor
```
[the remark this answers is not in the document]
kayibal: No, it's called BlockSynchronizer and lives in the client. I added this info here so people understand the parameters.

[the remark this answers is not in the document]
kayibal: Is it relevant? This is just to understand the parameters here, not to understand the tycho-client internals in detail - for this we should document what is mentioned here more detailed in tycho-client.

PR #349, opened 2025-09-30
feat: Decode error messages

tamaralipows:
- Also some useful implementations for parsing solidity errors.


tycho-integration-test/src/traces.rs
```diff
@@ -0,0 +1,356 @@
+//! Transaction trace analysis with foundry signature decoding.
```
tamaralipows, line 1: You need not review this file. It's exactly the same as in the testing SDK. Could we import it from somewhere shared? It sucks that it's two different repos.

tycho-integration-test/src/execution_simulator.rs
```diff
@@ -0,0 +1,496 @@
...
+        match code {
+            0x01 => "assertion failed".to_string(),
+            0x11 => "arithmetic underflow or overflow".to_string(),
+            0x12 => "division or modulo by zero".to_string(),
+            0x21 => "invalid enum value".to_string(),
+            0x22 => "storage byte array incorrectly encoded".to_string(),
+            0x31 => "pop on empty array".to_string(),
+            0x32 => "array index out of bounds".to_string(),
+            0x41 => "out of memory".to_string(),
+            0x51 => "invalid internal function".to_string(),
```
dianacarvalho1, line 369: oh nice

tycho-integration-test/src/four_byte_client.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub fn new() -> Self {
+        Self {
+            url: "https://www.4byte.directory/api/v1/signatures/".to_string(),
+            client: reqwest::Client::new(),
+        }
+    }
+
+    pub async fn get_signature(&self, hex_sig: &str) -> Result<String, Box<dyn std::error::Error>> {
+        info!("Fetching signature from 4Byte.directory");
+        let response = self
```
dianacarvalho1, line 37: uau you don't need any more weird things?? shocked. Why was I getting such weird signatures before with 4bytes? 🤔 was it because it was function signatures and not errors?

tycho-integration-test/src/main.rs
```diff
@@ -648,31 +641,65 @@ async fn setup_user_overwrites(
...
+        let allowance_storage_address = Address::from_slice(&allowance_storage_addr[..20]);
+
+        info!(
+            "Setting token override for {token_address}: balance={}, allowance={}, balance_storage={}, allowance_storage={}",
+            token_balance, token_allowance, balance_storage_address, allowance_storage_address
         );
+
+        // Apply balance and allowance overrides
+        // If both storage addresses are the same, combine them into one override
+        if balance_storage_address == allowance_storage_address {
```
dianacarvalho1, line 666: aaaahhhh they won't be the same when this is a proxy token??

dianacarvalho1: damn. we need to apply this fix in the sdk testing is well! Awesome job at figuring this out 👏🏼

tycho-integration-test/src/traces.rs
```diff
@@ -0,0 +1,356 @@
...
+        // Check for revert data in output (sometimes revert reasons are hex-encoded in output)
+        if let Some(output) = call_obj
+            .get("output")
+            .and_then(|v| v.as_str())
+        {
+            if !output.is_empty() && output != "0x" {
+                // Try to decode revert reason from output if it looks like revert data
+                if let Some(stripped) = output.strip_prefix("0x08c379a0") {
+                    // Error(string) selector
+                    if let Ok(decoded) = hex::decode(stripped) {
```
dianacarvalho1, line 320: should we try to use your error decoding here as well?

tycho-integration-test/src/traces.rs
dianacarvalho1: maybe we could have it all here and import from here in the sdk? 🤔

dianacarvalho1 (approved):
Thank you so much @tamaralipows 🙏🏼 Awesome work here! Only have small comments

tycho-integration-test/src/main.rs
```diff
@@ -172,6 +173,7 @@ async fn run(cli: Cli) -> miette::Result<()> {
                 continue;
             }
         };
+        // TODO why do we do this? Don't we also want to simulate on the first block?
```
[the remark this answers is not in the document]
dianacarvalho1: discussed this with @tamaralipows in a call. We should experiment in a separate PR if we can simulate in the first block or if it's too overwhelming

PR #352, opened 2025-10-01
fix: processing of bebop's aggregate multihop responses

adrianbenavides:
Instead of summing all the amounts in the response, only use those that match the tokens we are interested in.

I added a couple of tests:
- test_process_bebop_quote_response_aggregate_order: this was passing before the changes too. I added it just to verify I didn't change the existing behavior for regular aggregate responses
- test_process_bebop_quote_response_aggregate_order_with_multihop: this one failed with the current logic, now we get the expected amounts in/out.

### DIFF
--- a/.github/workflows/v2-main-workflow.yaml
+++ b/.github/workflows/v2-main-workflow.yaml
@@ -40,3 +40,21 @@ jobs:
     secrets:
       app_id: ${{ secrets.APP_ID }}
       app_private_key: ${{ secrets.APP_PRIVATE_KEY }}
+
+  build-and-push-integration-test:
+    needs:
+      - release
+    uses: propeller-heads/ci-cd-templates/.github/workflows/build-and-push-docker-image.yaml@main
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      image_tag: ${{ needs.release.outputs.next_release_version }}
+      image_name: tycho-integration-test
+    secrets:
+      app_id: ${{ secrets.APP_ID }}
+      app_private_key: ${{ secrets.APP_PRIVATE_KEY }}
+      role_to_assume: ${{ secrets.ROLE_TO_ASSUME }}
+      aws_region: ${{ secrets.AWS_REGION }}
+      repository_url: ${{ secrets.REPOSITORY_URL }}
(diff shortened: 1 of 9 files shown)


src/rfq/protocols/bebop/client.rs
```diff
@@ -149,6 +149,117 @@ impl BebopClient {
...
+                    },
+                    BebopOrderToSign::Aggregate(aggregate) => {
+                        // Sum taker_amounts for taker_tokens ending with the token_in
+                        let amount_in = aggregate
+                            .taker_tokens
+                            .iter()
+                            .enumerate()
+                            .filter(|(_, tokens)| {
+                                tokens
+                                    .last()
```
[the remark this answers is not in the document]
dianacarvalho1: huuumm but for the taker tokens, shouldn't it be the first one?? I'm confused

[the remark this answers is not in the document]
adrianbenavides: Yes, correct. This was wrong for the amount_in calculation. I'm working on a different approach to just store the matching indexes.

src/rfq/protocols/bebop/client.rs
```diff
@@ -149,6 +149,111 @@ impl BebopClient {
...
+                                                .get(taker_idx)
+                                                .and_then(|amounts| amounts.get(token_idx))
+                                                .and_then(|amount| BigUint::from_str(amount).ok())
+                                        } else {
+                                            None
+                                        }
+                                    })
+                                    .reduce(|a, b| a + b)
+                            })
+                            .sum();
```
dianacarvalho1, line 213: I think the logic is correct! ✅  this is gigantic though 😅 maybe we can use zip to aggregate tokens and amounts? 🤔

adrianbenavides: mmh good idea! Let me have a look

adrianbenavides: done! looks much better now 🙏

dianacarvalho1 (approved):
Amazing!! ✨  Thank you so much for fixing this so quickly @adrianbenavides 🙏🏼

PR #355, opened 2025-10-02
feat: Collect Integration Test Metrics

tamaralipows:
- Prometheus is the important thing for sending info to grafana
- Actix is just for exposing the endpoint locally - not sure if we need this
- I'm not sure we need opentelemetry at all for metrics - this might just be for tracing/collecting logs? Didn't fully understand yet - will look more tomorrow.


tycho-integration-test/src/main.rs
```diff
@@ -99,9 +101,22 @@ impl Debug for Cli {
...
+    // Initialize tracing with OpenTelemetry if configured
+    if let Ok(otlp_exporter_endpoint) = std::env::var("OTLP_EXPORTER_ENDPOINT") {
+        let config = ot::TracingConfig { otlp_exporter_endpoint };
+        ot::init_tracing(config)
+            .wrap_err("Failed to initialize OpenTelemetry tracing")?;
+    } else {
+        tracing_subscriber::fmt()
+            .with_env_filter(EnvFilter::from_default_env())
+            .init();
+    }
```
adrianbenavides, line 114: Since tracing is independent from telemetry, maybe this block and the ot::init_tracing function should be moved to a "tracing" module.

tycho-integration-test/src/metrics.rs
```diff
@@ -0,0 +1,63 @@
...
+            )
+        })
+        .bind(("0.0.0.0", 9898))
+        .expect("Failed to bind metrics server")
+        .run()
+        .await
+        {
+            error!("Metrics server failed: {}", e);
+        }
+    })
```
adrianbenavides, line 54: This will silently fail if the port is in use and metrics will get lost. We could return the task handle and await for it in parallel to the main task (e.g. with tokio::select).

We might also want to make the port configurable.

tycho-integration-test/src/ot.rs
```diff
@@ -0,0 +1,62 @@
...
+/// exporting trace data.
+pub fn init_tracing(config: TracingConfig) -> Result<()> {
+    global::set_text_map_propagator(TraceContextPropagator::new());
+
+    global::set_error_handler(|error| error!(error = format!("{error:#}"), "otel error"))
+        .map_err(|e| miette!("Failed to set error handler: {e}"))?;
+
+    let format = tracing_subscriber::fmt::format()
+        .with_level(true)
+        .with_target(false)
```
adrianbenavides, line 27: Any context on why we do this in our crates? It's impossible to see where a log line is coming from with this setting.

tycho-integration-test/Cargo.toml
```diff
@@ -21,7 +21,10 @@ serde_json = "1.0"
 tokio = { version = "1.47.1", features = ["full"] }
 tokio-retry2 = { version = "0.6.0", features = ["jitter"] }
 tracing = "0.1.41"
-tracing-subscriber = "0.3.20"
+tracing-subscriber = { version = "0.3.20", features = ["env-filter", "fmt"] }
+metrics = "0.24"
+metrics-exporter-prometheus = "0.16"
+actix-web = "4.4.0"
```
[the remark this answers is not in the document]
tamaralipows: Not the `tracing-subscriber` though!

PR #356, opened 2025-10-02
fix: Separate build + push integration test workflow from main workflow

dianacarvalho1:
This is so that the build workflow can be triggered manually and this way we don't need to merge the PR to main every time we want to test something


adrianbenavides (approved):
nice!

PR #358, opened 2025-10-02
feat: add support for RFQ protocols in integration test

adrianbenavides:
This pull request introduces a major refactor to the stream processing logic, adding dedicated processors for protocol and RFQ streams, improving error handling, and enhancing configuration and sampling.

The biggest change is that the `main.rs` has been split into two blocks: stream processing and swap simulation. The `stream_processor` module handles the logic of setting up and consuming the streams. In the `main.rs` file we now have a shared loop that consumes messages from either stream. Within this loop, most of the simulation logic is shared.

### DIFF
--- a/tycho-integration-test/src/main.rs
+++ b/tycho-integration-test/src/main.rs
@@ -27,7 +27,7 @@ use tracing_subscriber::EnvFilter;
 use tycho_common::simulation::protocol_sim::ProtocolSim;
 use tycho_simulation::{
     protocol::models::ProtocolComponent,
-    rfq::protocols::hashflow::client::HashflowClient,
+    rfq::protocols::hashflow::{client::HashflowClient, state::HashflowState},
     tycho_common::models::Chain,
     utils::{get_default_url, load_all_tokens},
 };
@@ -271,28 +271,30 @@ async fn process_update_state(
         return;
     }
     // Get all the possible swap directions
-    // For protocols supporting only one direction, use the first and last tokens
-    let protocols_supporting_one_direction_only = [HashflowClient::PROTOCOL_SYSTEM];
-    let swap_directions =
-        if protocols_supporting_one_direction_only.contains(&component.protocol_system.as_str()) {
-            // Return the first and the last tokens
-            vec![(
-                // We can safely unwrap here because we checked there are at least 2 tokens
-                component
-                    .tokens
-                    .first()
-                    .unwrap()
-                    .clone(),
-                component.tokens.last().unwrap().clone(),
-            )]
-        } else {
-            component
-                .tokens
-                .iter()
-                .permutations(2)
-                .map(|perm| (perm[0].clone(), perm[1].clone()))
-                .collect()
-        };
+    let swap_directions = match component.protocol_system.as_str() {
+        HashflowClient::PROTOCOL_SYSTEM => {
+            // Hashflow only supports swaps between the requested base and quote tokens
+            // WARN: we read from state because the component.tokens original order
+            // is modified here: src/protocol/models.rs: ProtocolComponent::from_with_tokens
+            let state = match state
+                .as_any()
+                .downcast_ref::<HashflowState>()
+            {
+                Some(s) => s.clone(),
+                None => {
+                    warn!("Failed to downcast state to HashflowState");
+                    return;
+                }
+            };
+            vec![(state.base_token, state.quote_token)]
+        }
+        _ => component
+            .tokens
+            .iter()
+            .permutations(2)
+            .map(|perm| (perm[0].clone(), perm[1].clone()))
+            .collect(),
+    };
     for (token_in, token_out) in swap_directions.iter() {
         info!(
             "Processing {} pool {state_id}, from {} to {}",


adrianbenavides:
I've merged the changes from https://github.com/propeller-heads/tycho-simulation/pull/349/ and fixed all the merge conflicts.

tycho-integration-test/src/rfq_stream_processor.rs
```diff
@@ -0,0 +1,173 @@
...
+                }
+            }
+        }
+
+        // Start the RFQ stream
+        let mut is_first_update = false;
+        let (tx, mut rx) = tokio::sync::mpsc::channel(64);
+        let _handle = tokio::spawn(rfq_stream_builder.build(tx));
+        let sample_size = self.sample_size;
+        let stream_sleep = self.stream_sleep;
```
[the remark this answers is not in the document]
dianacarvalho1: yes I agree with skipping messages. I was thinking about this yesterday and I think it would be actually ok to just test the RFQs every ten minutes or so. No need to do it sooooo often. It will be really overwhelming and kinda redundant

tycho-integration-test/src/main.rs
```diff
@@ -160,669 +126,301 @@ async fn run(cli: Cli) -> miette::Result<()> {
...
+        tycho_url.clone(),
+        cli.tycho_api_key.clone(),
+        cli.tvl_threshold,
+    ) {
+        protocol_stream_processor
+            .run_stream(&all_tokens, tx.clone())
+            .await?;
+    }
+    if let Ok(rfq_stream_processor) =
+        RfqStreamProcessor::new(chain, cli.tvl_threshold, cli.max_n_simulations)
```
tamaralipows, line 151: What happens if one of them fails? Are we throwing away the error? Shouldn't we bubble it up instead?

tycho-integration-test/src/stream_processor/protocol_stream_processor.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    pub async fn run_stream(
+        &self,
+        all_tokens: &HashMap<Bytes, Token>,
+        tx: Sender<StreamUpdate>,
+    ) -> miette::Result<JoinHandle<()>> {
+        let mut stream = self.build_stream(all_tokens).await?;
+        let handle = tokio::spawn(async move {
+            let mut is_first_update = false;
```
dianacarvalho1, line 58: shouldn't this be true?

tycho-integration-test/src/stream_processor/rfq_stream_processor.rs
```diff
@@ -0,0 +1,174 @@
...
+            info!("Bebop RFQ credentials found.");
+            rfq_credentials.insert(RfqProtocol::Bebop, (user, key));
+        }
+        let (hashflow_user, hashflow_key) =
+            (env::var("HASHFLOW_USER").ok(), env::var("HASHFLOW_KEY").ok());
+        if let (Some(user), Some(key)) = (hashflow_user, hashflow_key) {
+            info!("Hashflow RFQ credentials found.");
+            rfq_credentials.insert(RfqProtocol::Hashflow, (user, key));
+        }
+        if rfq_credentials.is_empty() {
```
dianacarvalho1, line 63: this error is only raised if both credentials are missing, if one is there but the other is not this will silently continue

tycho-integration-test/src/stream_processor/rfq_stream_processor.rs
```diff
@@ -0,0 +1,174 @@
...
+                    .collect();
+                update
+                    .new_pairs
+                    .retain(|key, _| update.states.contains_key(key));
+
+                // Send the latest update
+                let update = StreamUpdate { update_type: UpdateType::Rfq, update, is_first_update };
+                if is_first_update {
+                    is_first_update = false;
+                }
```
dianacarvalho1, line 164: it will always be false here

dianacarvalho1 (commented):
Thank you! I have a few more questions..

tycho-integration-test/src/stream_processor/rfq_stream_processor.rs
```diff
@@ -68,7 +68,7 @@ impl RfqStreamProcessor {
             tvl_threshold,
             rfq_credentials,
             sample_size: 10,
-            stream_sleep_time: Duration::from_secs(10 * 60),
+            stream_sleep_time: Duration::from_secs(60),
```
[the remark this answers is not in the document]
adrianbenavides: I'll add an arg in the Cli for this

tycho-integration-test/src/stream_processor/rfq_stream_processor.rs
adrianbenavides: That's intentional:
- It will only run the configured protocols, but the binary doesn't require to pass any. Maybe you want to run this binary for regular protocols only.
- If none are provided, then we return an error and we don't start the stream [here](https://github.com/propeller-heads/tycho-simulation/blob/d94ee343b5917020bc3e60fffe8ce384612901a3/tycho-integration-test/src/main.rs#L153)

tycho-integration-test/src/stream_processor/rfq_stream_processor.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+}
+
+pub struct RfqStreamProcessor {
+    chain: Chain,
+    tvl_threshold: f64,
+    rfq_credentials: HashMap<RfqProtocol, (String, String)>,
+    sample_size: usize,
+    stream_sleep_time: Duration,
```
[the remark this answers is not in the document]
adrianbenavides: I'll rename it to skip_messages_duration and add a comment.

PR #557, opened 2026-03-18
feat: native Curve StableSwap 2-token pool simulation

sunce86:
## Summary

Native Rust implementation of the Curve StableSwap invariant for 2-token plain pools (e.g., DAI/USDC, FRAX/USDC, USDS/USDC). This replaces the VM/revm fallback for these pool types, providing ~100x faster swap simulation.

## Motivation

StableSwap pools are among the highest-volume DeFi pools. The current VM fallback executes Curve bytecode in revm at ~100-200μs per simulation. A native implementation runs at ~900ns (benchmarked in a separate project), enabling significantly better routing performance when evaluating thousands of pool candidates.

## Implementation

**Math (`math.rs`):**
- `get_d()` — Newton-Raphson solver for StableSwap invariant D
- `get_y()` — Newton-Raphson solver for output balance given input and D
- `spot_price_raw()` — Analytical marginal price from invariant derivative (no simulation needed)
- All arithmetic ported line-by-line from [Curve's Vyper source](https://github.com/curvefi/curve-contract/blob/master/contracts/pool-templates/base/SwapTemplateBase.vy), matching multiplication order for bit-for-bit identical results

**State (`state.rs`):**
- `ProtocolSim` implementation with `get_amount_out`, `spot_price`, `get_limits`, `delta_transition`
- Reserves stored in raw (native) token decimals; normalized to 18 decimals internally via rates (matching Curve's RATES mechanism)
- Handles mixed-decimal pairs (e.g., DAI 18-dec / USDC 6-dec)

**Decoder (`decoder.rs`):**
- Decodes `reserve0`, `reserve1`, `A`, `fee` from indexer attributes
- Extracts token decimals from `all_tokens` for rate computation
- Converts raw `A()` to internal precision (`A * A_PRECISION`)

## Verification

**On-chain verification** against FRAX/USDC pool (`0xDcEF968d416a41Cdac0ED8702fAC8128A64241A2`) on Ethereum mainnet:
- 6 swap amounts tested (both directions: FRAX→USDC and USDC→FRAX)
- All `assert_eq!` — **bit-for-bit identical** to on-chain `get_dy()`, zero difference

**30 unit tests** covering:
- Math convergence (balanced, imbalanced, high/low A, symmetry)
- Spot price (balanced pool, imbalanced, inverse property)
- Swap amounts (same decimals, mixed decimals, both directions, imbalanced)
- Delta transitions (reserves, amplification, missing attributes)
- Limits (both directions, empty pool, max_out ≤ reserve)
- Roundtrip consistency, state immutability

## Usage

```rust
.exchange::<CurveStableSwapState>("curve_stableswap", filter, None)
```

## Scope & Limitations

- **2-token plain pools only** — MetaPools and Tricrypto are separate invariants
- **Requires parsed attributes** from indexer (`reserve0`, `reserve1`, `A`, `fee`). The current Curve substreams module may send raw storage slots for the VM adapter instead. An indexer update to emit parsed attributes would be needed.
- `query_pool_swap` not yet implemented (returns error, consistent with V4/VM)

## Follow-up

- MetaPool support (nested pool resolution)
- N-token plain pool generalization
- `query_pool_swap` implementation for target-price solving

### DIFF
--- a/src/evm/protocol/curve_stableswap/decoder.rs
+++ b/src/evm/protocol/curve_stableswap/decoder.rs
@@ -1,0 +1,220 @@
+use std::collections::HashMap;
+
+use alloy::primitives::U256;
+use tycho_client::feed::{synchronizer::ComponentWithState, BlockHeader};
+use tycho_common::{models::token::Token, Bytes};
+
+use super::{math, state::CurveStableSwapState};
+use crate::protocol::{
+    errors::InvalidSnapshotError,
+    models::{DecoderContext, TryFromWithBlock},
+};
+
+impl TryFromWithBlock<ComponentWithState, BlockHeader> for CurveStableSwapState {
+    type Error = InvalidSnapshotError;
+
+    async fn try_from_with_header(
+        snapshot: ComponentWithState,
+        _block: BlockHeader,
+        _account_balances: &HashMap<Bytes, HashMap<Bytes, Bytes>>,
+        all_tokens: &HashMap<Bytes, Token>,
+        _decoder_context: &DecoderContext,
+    ) -> Result<Self, Self::Error> {
+        let attributes = &snapshot.state.attributes;
+
+        let reserve0 = U256::from_be_slice(
+            attributes
+                .get("reserve0")
+                .ok_or(InvalidSnapshotError::MissingAttribute("reserve0".to_string()))?,
+        );
+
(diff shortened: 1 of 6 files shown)


kayibal:
@sunce86 Sorry for the late reply, but yes we would be very interested in such a port.

PR #566, opened 2026-04-01
feat: add missing components to the blocklist

Troshchk:
- Add built-in default blocklist used as a fallback when no config is provided
- Update blocklist with missing components

### DIFF
--- a/blocklist.toml
+++ b/blocklist.toml
@@ -3,14 +3,29 @@
 
 [blocklist]
 # Component IDs to exclude (pool addresses)
-# AMPL pools - AMPL is a rebasing token that breaks simulation assumptions
-# UniswapV3 AMPL/WETH
-# UniswapV2 AMPL/WETH
 components = [
+    # AMPL pools - AMPL is a rebasing token that breaks simulation assumptions
+    # UniswapV3 AMPL/WETH
+    # UniswapV2 AMPL/WETH
     "0x86d257cdb7bc9c0df10e84c8709697f92770b335",
     "0xc5be99a02c6857f9eac67bbce58df5572498f40c",
     # Fluid Lite pools with broken simulation (ENG-5696)
     # Off-chain sim claims 50-100x more output than on-chain reality
     "0x32aa6f5c6f771b39d383c6e36d7ef0702a28e32ad64671c059f41547b82ef0ef",
     "0x7f31b44f032f125bb465c161343fccfdc88fee8dc94c068f6430d2345d80f1d1",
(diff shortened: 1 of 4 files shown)


zizou0x (approved):
LGTM! Thanks!

blocklist.toml
```diff
@@ -3,14 +3,29 @@
...
     # Fluid Lite pools with broken simulation (ENG-5696)
     # Off-chain sim claims 50-100x more output than on-chain reality
     "0x32aa6f5c6f771b39d383c6e36d7ef0702a28e32ad64671c059f41547b82ef0ef",
     "0x7f31b44f032f125bb465c161343fccfdc88fee8dc94c068f6430d2345d80f1d1",
+    # Curve yETH/WETH — exploited Nov 2025 ($9M), broken invariant, pool insolvent
+    "0x69accb968b19a53790f43e57558f5e443a91af22",
+    # Fluid syrupUSDC/USDC — ERC-4626 vault token, simulation can't track accumulating rate
+    "0x79eea4a1be86c43a9a9c4384b0b28a07af24ae29",
+    # Curve weETH/WETH (StableSwapNG) — MissingAccount error, oracle deps not in simulated state
+    "0xdb74dfdd3bb46be8ce6c33dc9d82777bcfc3ded5",
```
zizou0x: This is recoverable and doesn't affect simulation, I wouldn't have it here

src/evm/stream.rs
```diff
@@ -365,16 +367,16 @@ impl ProtocolStreamBuilder {
...
     ///
-    /// Loads blocklisted component IDs from a TOML file at `path`.
-    /// If the file doesn't exist or can't be parsed, no components are
-    /// excluded and a warning is logged.
-    pub fn blocklist_components(mut self, path: impl AsRef<Path>) -> Self {
-        let ids = crate::utils::load_blocklist(path.as_ref());
+    /// If `path` is `Some`, loads the blocklist from that file and
+    /// returns an error if missing or invalid. If `None`, uses the
+    /// default blocklist embedded in the library.
+    pub fn blocklist_components(mut self, path: Option<&Path>) -> Result<Self, SimulationError> {
```
kayibal: Wouldn't it be better to pass a list here instead of a file path? This way the method can't fail.

kayibal (changes requested):

src/evm/stream.rs
Troshchk: Changed

kayibal (approved):

PR #567, opened 2026-04-03
feat: aerodrome v1 simulation

zach030:


src/evm/protocol/aerodrome_v1/decoder.rs
```diff
@@ -0,0 +1,268 @@
...
+            Some(&1);
+
+        let fee = snapshot
+            .state
+            .attributes
+            .get("fee")
+            .map(|fee| u32::from(fee.clone()))
+            .unwrap_or(0);
+
+        if fee > 10_000 && fee != 420 {
```
zizou0x: The second condition is redundant and will always be true if the first is true

src/evm/protocol/mod.rs
```diff
@@ -11,6 +12,7 @@ pub mod fluid;
 pub mod pancakeswap_v2;
 pub mod rocketpool;
 pub mod safe_math;
+mod solidly_stable;
```
zizou0x: Are we foreseeing any future usage of these? If we do then we can keep them here ready to be re-used, else I'd have them into aerodrome_v1 directly because there is a high chance we aren't gonna re-use them ever.

zizou0x (approved):
LGTM! thanks!

